PR #8, opened 2020-04-22
DM-16708 - Better dry-run support

brianv0:
The docker container has been used for quite a while, but this code hadn't been merged to master.


brianv0:
I tested this out and the quoting works fine in the sdm_schema/yml files which. This doesn't capture other input errors in those yaml files, however, but the quoting seems fine.

PR #12, opened 2021-04-28
Fix table_index to be from table, not schema

brianv0:


gpdf (commented):
Looks right to me.  I also checked that `schema_index` and `column_index` are handled self-consistently following the same pattern.

PR #13, opened 2021-06-10
Add table name overrides to TAP schema names. Implement felis file merging.

brianv0:
This adds a few different options for overriding the names of the TAP tables, either individually or with a postfix option (`--tap-tables-postfix`).

This also fixes a few issues around felis file merging as well provides an small example for merging.

### DIFF
--- a/examples/merge/point-extra.yml
+++ b/examples/merge/point-extra.yml
@@ -1,0 +1,8 @@
+---
+"@graph":
+  - "@id": "#Geometry_Point.ra"
+    ivoa:ucd: pos.eq.ra;meta.main
+    fits:tunit: deg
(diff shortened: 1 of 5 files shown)


gpdf (commented):
I'm not really able to review the code changes in detail, since I'm not familiar with the existing internals, but superficially they seem to address a large part of the current need.

In particular, we can now generate the "tables11"-style table names automatically, as well as apply overrides to specific `schema_index` values, for purposes of curating the presentation order.

Still missing is the ability to globally rewrite `schema_index` for `schema_index=0` and `schema_index`-unset cases, to some command-line-specified large value (to push them beyond the known/curated ones).  But it would be fine to merge what is already on this PR and get the downstream automation established.

PR #18, opened 2023-02-10
DM-37744: Make felis compatible with sqlalchemy 2

andy-slac:
Few small changes to make it work with both 1.4 and 2.0, plus mypy updates for new sqlalchemy.


timj (approved):
Looks good. Can you also please run `pre-commit autoupdate` to update the pre-commit configuration? (and maybe add a flake8 section to it copied from daf_butler)

PR #19, opened 2023-03-09
DM-38304: Fix crash in `load-tap --dry-run` command

andy-slac:
`create_mock_engine` does not actually return an Engine, but a MockConnection instance which has very different API. Updated TapLoadingVisitor to handle their differences correctly. Added unit test for `load-tap --dry-run`.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -157,13 +157,6 @@ def load_tap(
     if isinstance(normalized["@graph"], dict):
         normalized["@graph"] = [normalized["@graph"]]
 
-    if not dry_run:
-        engine = create_engine(engine_url)
-    else:
-        _insert_dump = InsertDump()
-        engine = create_engine(engine_url, strategy="mock", executor=_insert_dump.dump, paramstyle="pyformat")
-        # After the engine is created, update the executor with the dialect
-        _insert_dump.dialect = engine.dialect
     tap_tables = init_tables(
         tap_schema_name,
         tap_tables_postfix,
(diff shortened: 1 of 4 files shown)


timj:
Can you turn on python 3.11 in the build matrix?

python/felis/cli.py
```diff
@@ -174,15 +167,35 @@ def load_tap(
...
-        tap_visitor = TapLoadingVisitor(
-            engine, catalog_name=catalog_name, schema_name=schema_name, mock=dry_run, tap_tables=tap_tables
-        )
-        tap_visitor.visit_schema(schema)
+        if engine_url == "sqlite://" and not dry_run:
+            # In Memory SQLite - Mostly used to test
+            Tap11Base.metadata.create_all(engine)
+
+        for schema in normalized["@graph"]:
+            tap_visitor = TapLoadingVisitor(
```
timj, line 178: I suppose we could store `TapLoadingVisitor` or `TapLoadingVisit.from_mock_connection` in a variable and then have this loop outside the `if`-blocks so it's not repeated code, but it's probably not worth it.

timj (approved):
Looks okay. Please add python 3.11 to matrix.

python/felis/cli.py
andy-slac: I thought about moving it out of the loop, but it looks like there is some accumulated internal state in the visitor, so it may not be 100% safe to do it. Maybe something to cleanup later.

PR #23, opened 2023-09-14
DM-40675: Implement schema version specification for RFC-954

andy-slac:
README defines the structure for `version` keyword that can be either a
simple string or a structure specifying version compatibility. Visitor
classes are extended to handle version information, for now only the
`SimpleVisitor` converts version info into a usable structure.

Plus some code cleanup and ruff checks for docstrings.


.pre-commit-config.yaml
```diff
@@ -7,8 +7,9 @@ repos:
           - "--unsafe"
       - id: end-of-file-fixer
       - id: trailing-whitespace
+        exclude: "README.md"
   - repo: https://github.com/psf/black
```
timj, line 11: Note that black now has a specialist repo for pre-commit at https://github.com/psf/black-pre-commit-mirror

.pre-commit-config.yaml
```diff
@@ -22,6 +23,6 @@ repos:
       - id: isort
         name: isort (python)
   - repo: https://github.com/PyCQA/flake8
```
timj, line 25: Use ruff in pre-commit?

.pre-commit-config.yaml
```diff
@@ -7,21 +7,23 @@ repos:
           - "--unsafe"
       - id: end-of-file-fixer
       - id: trailing-whitespace
-  - repo: https://github.com/psf/black
-    rev: 23.1.0
+        exclude: "README.md"
```
timj, line 10: I'm very confused.  Why can't we remove the trailing whitespace from README.md?

README.md
```diff
@@ -306,12 +308,50 @@ may include:
...
+relations but does not specify how exactly compatibility checks have to be
+implemented. It is the client responsibility to interpret version numbers and
+to define compatibility rules.
+
+In simplest form the schema version can be specified as a value for the
+`version` attribute and it must be a string:
+
+    version: "4.2.0"
+
+This example uses semantic version format, but in general any string or number
```
timj, line 330: Maybe we need to note that semantic version is not relevant at all because even with semantic versioning you will need to list every version (including patchlevel) in the compatible and read_compatible sections as needed and nothing can be implied by semantics. (you could imagine that minor version change meant read_compatible and patchlevel version changes were fully compatible if you wanted to enforce semantics)

doc/conf.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+
+from documenteer.conf.pipelinespkg import *  # noqa: F403, import *
+
+project = "felis"
+html_theme_options["logotext"] = project  # noqa: F405, unknown name
+html_title = project
+html_short_title = project
+doxylink = {}
+exclude_patterns = ["changes/*"]
```
timj, line 13: Maybe I should encourage the new data engineer to start using towncrier when they start developing this package.

doc/index.rst
```diff
@@ -276,18 +276,45 @@ tables. The core annotations of a schema are:
...
 
 -  DBMS-specific information for a schema, especially for creating a schema.
 -  IVOA metadata about the table
--  Schema versioning information
 -  Column Groupings
 -  Links to other schemas which may be related
 -  Reference URLs
 -  Provenance
 
+Schema versioning
```
timj, line 291: Now I'm wondering how much information we really want to duplicate in the sphinx docs and the README, especially given that the readme is in markdown and not rst. Sphinx should be primary, although at the very least we could change to `README.rst` so that at least it would be a simple copy and paste, not reformatting. Something for next time.

python/felis/check.py
```diff
@@ -67,6 +67,53 @@ def check_schema(self, schema_obj: _Mapping) -> None:
         _id = self._assert_id(schema_obj)
         self._check_visited(_id)
 
+    def check_schema_version(self, version_obj: Any, schema_obj: _Mapping) -> None:
+        """Validate contents of schema description object.
+
+        Parameters
+        ----------
+        version_obj : `Any`
```
timj, line 75: Confused about the use of `Any` here when we are informed that it's either a string or a Mapping. I assume it's because the caller is using dict where it doesn't know any of the value types.

python/felis/check.py
```diff
@@ -67,6 +67,53 @@ def check_schema(self, schema_obj: _Mapping) -> None:
...
+            if not isinstance(version_obj["current"], str):
+                raise TypeError(f"schema version 'current' value is not a string: {version_obj['current']!r}")
+            if (compatible := version_obj.get("compatible")) is not None:
+                if not isinstance(compatible, list):
+                    raise TypeError(f"schema version 'compatible' value is not a list: {compatible!r}")
+                for item in compatible:
+                    if not isinstance(item, str):
+                        raise TypeError(f"items in 'compatible' value are not strings: {compatible!r}")
+            if (read_compatible := version_obj.get("read_compatible")) is not None:
+                if not isinstance(read_compatible, list):
```
timj, line 105: All this code makes me dream of felis models being defined with Pydantic...

python/felis/simple.py
```diff
@@ -211,6 +212,21 @@ class Table:
...
 
 
+@dataclasses.dataclass
+class SchemaVersion:
+    """Schema versioning description."""
+
+    current: str
+    """Current schema version defined by the document."""
+
+    compatible: Optional[List[str]] = None
```
timj, line 222:
```suggestion
    compatible: list[str] | None = None
```

Can you modernize all the type annotations? pyupgrade or `ruff --select UP --fix` should manage it for you.

timj (approved):
Looks okay. All the manual model validation seems painful to me in a world where Pydantic exists. Please modernize type annotations.

.pre-commit-config.yaml
andy-slac: Markdown treats two consecutive spaces as `<b/>` and this is used in README to generate nicely formatted output. I think we'll want to drop `README` in the future because the same contents is in `doc/index.rst`. I'll make a ticket for myself to improve documentation.

timj: I'm pretty sure GitHub markdown allows a `\` at the end of the line to be equivalent to the trailing spaces.

README.md
andy-slac: Well, semantic version can be relevant if you can define compatibility rules based on that. `daf_butler` uses semantic versioning (mostly for code compatibility, but it also implies schema compatibility). My idea is that if you can use semantic versioning then you don't have to use `compatible`/`read_compatible` and can just compare version in YAML file with the version that is stored in the database.

python/felis/check.py
andy-slac: Here it checks whatever comes from YAML and that can be an arbitrary thing. Only after validation we can be sure that types are correct.

README.md
timj: Ok. It wasn't clear to me that semantic could be special as opposed to requiring every compatible string version to be listed.

python/felis/check.py
andy-slac: I thought about Pydantic, but did not want to introduce the dependency just yet, there is a lot of code in this package that I still don't quite understand. We'll need another ticket to make use of Pydantic.

python/felis/simple.py
andy-slac: We don't need `3.9` for this package, right? (it does not like `type[...]`)

timj: Correct. 3.11 is fine but keeping 3.10 around is okay.

andy-slac: @timj, GA still wants to build 3.9, I guess because it's `required`. I don't remember how to remove that and I probably don't have permissions for this package, could you clean that?

timj: Done.

.pre-commit-config.yaml
andy-slac: I switched to `\`.

PR #24, opened 2023-10-18
DM-41218: Print detailed error messages when checking IVOA UCD syntax

JeremyMcCormick:
This replaces usage of the astropy check_ucd function with parse_ucd, catches any errors that occur, and prints them to the error log.

### DIFF
--- a/python/felis/check.py
+++ b/python/felis/check.py
@@ -159,8 +159,11 @@ class FelisValidator:
 
         # Check UCDs of columns
         ivoa_ucd = column_obj.get("ivoa:ucd")
-        if not ucd.check_ucd(ivoa_ucd, check_controlled_vocabulary=True):
-            logger.error(f"invalid ucd for {_id}: {ivoa_ucd}")
+        if ivoa_ucd:
+            try:
+                ucd.parse_ucd(ivoa_ucd, check_controlled_vocabulary=True, has_colon=";" in ivoa_ucd)
+            except ValueError as e:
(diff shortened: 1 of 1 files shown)


python/felis/check.py
```diff
@@ -159,8 +159,11 @@ def check_column(self, column_obj: _Mapping, table_obj: _Mapping) -> None:
 
         # Check UCDs of columns
         ivoa_ucd = column_obj.get("ivoa:ucd")
-        if not ucd.check_ucd(ivoa_ucd, check_controlled_vocabulary=True):
-            logger.error(f"invalid ucd for {_id}: {ivoa_ucd}")
+        if ivoa_ucd:
+            try:
+                ucd.parse_ucd(ivoa_ucd, check_controlled_vocabulary=True, has_colon=";" in ivoa_ucd)
+            except ValueError as e:
+                logger.error(f"{e} in '{ivoa_ucd}' for '{_id}'")
```
gpdf, line 166: Let's say 

... in UCD '{ivoa_ucd}' ...

Especially since the error messages from `parse_ucd` don't themselves use the term.

gpdf (approved):
That looks good, and meets the spec for this change, though do please make the message change.  

In a future meeting, let's take some time to figure out, for a subsequent ticket, how we will distinguish between "I haven't bothered to set a UCD for column X" and "I've studied the matter and there is no appropriate UCD for column X".

gpdf (approved):
Looks perfect now.

PR #25, opened 2023-11-21
DM-41688: Add Pydantic data model to felis

JeremyMcCormick:
This PR adds a Pydantic v2 data model to felis describing the YAML schema format, as well as unit tests for the new module. A `validate` command was added to the CLI which will load any number of YAML files into Pydantic and print validation errors. This has been tested on all of the schemas in `sdm_schemas/yml`.

To run tests covering the new module:

```
cd tests && python -m unittest test_datamodel
```

To run validation on all of the SDM schemas using the CLI tool:

```
# Change to use your sdm_schemas location
felis validate ../sdm_schemas/yml/*.yaml
```

These changes make no attempt to replace usage of the existing "simple" model that is used throughout Felis. All existing functionality should be unaffected.

For more details, please see the message on this PR's single commit. Updates to the Felis documentation will be forthcoming.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -150,12 +150,12 @@ def load_tap(
         schema_obj = {"@context": DEFAULT_CONTEXT, "@graph": top_level_object}
     else:
         logger.error("Schema object not of recognizable type")
-        sys.exit(1)
+        raise click.exceptions.Exit(1)
 
     normalized = _normalize(schema_obj, embed="@always")
     if len(normalized["@graph"]) > 1 and (schema_name or catalog_name):
         logger.error("--schema-name and --catalog-name incompatible with multiple schemas")
-        sys.exit(1)
+        raise click.exceptions.Exit(1)
 
     # Force normalized["@graph"] to a list, which is what happens when there's
     # multiple schemas
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -28,12 +28,14 @@ from typing import Any
 
 import click
 import yaml
+from pydantic import ValidationError
 from pyld import jsonld
 from sqlalchemy.engine import Engine, create_engine, create_mock_engine, make_url
 from sqlalchemy.engine.mock import MockConnection
 
 from . import DEFAULT_CONTEXT, DEFAULT_FRAME, __version__
 from .check import CheckingVisitor
+from .datamodel import Schema
 from .sql import SQLVisitor
 from .tap import Tap11Base, TapLoadingVisitor, init_tables
 from .utils import ReorderingVisitor
@@ -297,6 +299,23 @@ def merge(files: Iterable[io.TextIOBase]) -> None:
     merged = {"@context": DEFAULT_CONTEXT, "@graph": list(updated_map.values())}
     normalized = _normalize(merged, embed="@always")
     _dump(normalized)
+
+
+@cli.command("validate")
+@click.argument("files", nargs=-1, type=click.File())
+def validate(files: Iterable[io.TextIOBase]) -> None:
+    """Validate one or more felis YAML files."""
+    rc = 0
+    for file in files:
+        file_name = getattr(file, "name", None)
+        logger.info(f"Validating {file_name}")
+        try:
+            Schema.model_validate(yaml.load(file, Loader=yaml.SafeLoader))
+        except ValidationError as e:
+            logger.error(e)
+            rc = 1
+    if rc:
+        raise click.exceptions.Exit(rc)
 
 
 @cli.command("dump-json")
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -1,0 +1,356 @@
+# This file is part of felis.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import logging
+from collections.abc import Mapping
+from enum import Enum
+from typing import Any, Literal
+
+from astropy import units as units  # type: ignore
+from astropy.io.votable import ucd  # type: ignore
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+logger = logging.getLogger(__name__)
+# logger.setLevel(logging.DEBUG)
+
+
+class BaseObject(BaseModel):
+    """Base class for all Felis objects."""
+
+    model_config = ConfigDict(populate_by_name=True, extra="forbid", use_enum_values=True)
+    """Configuration for the `BaseModel` class.
+
+    Allow attributes to be populated by name and forbid extra attributes.
+    """
+
+    name: str
+    """The name of the database object.
+
+    All Felis database objects must have a name.
+    """
+
+    id: str = Field(alias="@id")
+    """The unique identifier of the database object.
+
+    All Felis database objects must have a unique identifier.
+    """
+
+    description: str | None = None
+    """A description of the database object.
+
+    The description is optional.
+    """
+
+
+class DataType(Enum):
+    """`Enum` representing the data types supported by Felis."""
+
+    BOOLEAN = "boolean"
+    BYTE = "byte"
+    SHORT = "short"
+    INT = "int"
+    LONG = "long"
+    FLOAT = "float"
+    DOUBLE = "double"
+    CHAR = "char"
+    STRING = "string"
+    UNICODE = "unicode"
+    TEXT = "text"
+    BINARY = "binary"
+    TIMESTAMP = "timestamp"
+
+
+class Column(BaseObject):
+    """A column in a table."""
+
+    datatype: DataType
+    """The datatype of the column."""
+
+    length: int | None = None
+    """The length of the column."""
+
+    nullable: bool = True
+    """Whether the column can be `NULL`."""
+
+    value: Any = None
+    """The default value of the column."""
+
+    autoincrement: bool | None = None
+    """Whether the column is autoincremented."""
+
+    mysql_datatype: str | None = Field(None, alias="mysql:datatype")
+    """The MySQL datatype of the column."""
+
+    ivoa_ucd: str | None = Field(None, alias="ivoa:ucd")
+    """The IVOA UCD of the column."""
+
+    fits_tunit: str | None = Field(None, alias="fits:tunit")
+    """The FITS TUNIT of the column."""
+
+    ivoa_unit: str | None = Field(None, alias="ivoa:unit")
+    """The IVOA unit of the column."""
+
+    tap_column_index: int | None = Field(None, alias="tap:column_index")
+    """The TAP_SCHEMA column index of the column."""
+
+    tap_principal: int | None = Field(0, alias="tap:principal", ge=0, le=1)
+    """Whether this is a TAP_SCHEMA principal column; can be either 0 or 1.
+
+    This could be a boolean instead of 0 or 1.
+    """
+
+    votable_arraysize: int | Literal["*"] | None = Field(None, alias="votable:arraysize")
+    """The VOTable arraysize of the column."""
+
+    tap_std: int | None = Field(0, alias="tap:std", ge=0, le=1)
+    """TAP_SCHEMA indication that this column is defined by an IVOA standard.
+    """
+
+    votable_utype: str | None = Field(None, alias="votable:utype")
+    """The VOTable utype (usage-specific or unique type) of the column."""
+
+    votable_xtype: str | None = Field(None, alias="votable:xtype")
+    """The VOTable xtype (extended type) of the column."""
+
+    @field_validator("ivoa_ucd")
+    @classmethod
+    def check_ivoa_ucd(cls, ivoa_ucd: str) -> str:
+        """Check that IVOA UCD values are valid."""
+        if ivoa_ucd is not None:
+            try:
+                ucd.parse_ucd(ivoa_ucd, check_controlled_vocabulary=True, has_colon=";" in ivoa_ucd)
+            except ValueError as e:
+                raise ValueError(f"Invalid IVOA UCD: {e}")
+        return ivoa_ucd
+
+    @model_validator(mode="before")
+    @classmethod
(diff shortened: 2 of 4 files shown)


JeremyMcCormick:
mypy warnings/errors:

```
python/felis/datamodel.py:26: error: Skipping analyzing "astropy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
python/felis/datamodel.py:27: error: Skipping analyzing "astropy.io.votable": module is installed, but missing library stubs or py.typed marker  [import-untyped]
python/felis/datamodel.py:27: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
python/felis/datamodel.py:133: error: Function is missing a return type annotation  [no-untyped-def]
python/felis/datamodel.py:144: error: Function is missing a type annotation  [no-untyped-def]
python/felis/datamodel.py:237: error: Function is missing a type annotation  [no-untyped-def]
python/felis/datamodel.py:251: error: Function is missing a type annotation  [no-untyped-def]
python/felis/datamodel.py:282: error: Function is missing a type annotation  [no-untyped-def]
python/felis/cli.py:310: error: "TextIOBase" has no attribute "name"  [attr-defined]
python/felis/cli.py:3[16](https://github.com/lsst/felis/actions/runs/6948155795/job/18903507945?pr=25#step:7:17): error: No return value expected  [return-value]
Found 9 errors in 2 files (checked 13 source files)
```

JeremyMcCormick:
TODO: Move to Pydantic v2 instead of using v1 methods, e.g. use model_validator instead of root_validator, etc.

JeremyMcCormick:
> mypy warnings/errors:
> 
> ```
> python/felis/datamodel.py:26: error: Skipping analyzing "astropy": module is installed, but missing library stubs or py.typed marker  [import-untyped]
> python/felis/datamodel.py:27: error: Skipping analyzing "astropy.io.votable": module is installed, but missing library stubs or py.typed marker  [import-untyped]
> python/felis/datamodel.py:27: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> python/felis/datamodel.py:133: error: Function is missing a return type annotation  [no-untyped-def]
> python/felis/datamodel.py:144: error: Function is missing a type annotation  [no-untyped-def]
> python/felis/datamodel.py:237: error: Function is missing a type annotation  [no-untyped-def]
> python/felis/datamodel.py:251: error: Function is missing a type annotation  [no-untyped-def]
> python/felis/datamodel.py:282: error: Function is missing a type annotation  [no-untyped-def]
> python/felis/cli.py:310: error: "TextIOBase" has no attribute "name"  [attr-defined]
> python/felis/cli.py:3[16](https://github.com/lsst/felis/actions/runs/6948155795/job/18903507945?pr=25#step:7:17): error: No return value expected  [return-value]
> Found 9 errors in 2 files (checked 13 source files)
> ```

These are fixed now.

JeremyMcCormick:
@timj suggests using type annotation style like: `str | None` instead of `Optional[str]` and fixing up using `ruff`.

JeremyMcCormick:
Sample output of new validation using Pydantic on `sdm_schemas/yml/*.yaml`:

[validate.log](https://github.com/lsst/felis/files/13431602/validate.log)

timj:
deescription? (at start of validate.log)

JeremyMcCormick:
> deescription? (at start of validate.log)

Someone misspelled "description" in the SDM schema itself, and the strict validation has caught the error. 

Before this would not have been caught because extra attributes were allowed in the YAML files and they are disallowed now.

python/felis/cli.py
```diff
@@ -299,6 +301,22 @@ def merge(files: Iterable[io.TextIOBase]) -> None:
...
+    rc = 0
+    for file in files:
+        file_name = getattr(file, "name", None)
+        logger.info(f"Validating {file_name}")
+        try:
+            Schema.model_validate(yaml.load(file, Loader=yaml.SafeLoader))
+        except ValidationError as e:
+            logger.error(e)
+            rc = 1
+    return sys.exit(rc)
```
timj, line 317:
```suggestion
    if rc:
        raise click.exception.Exit(rc)
```
I think you are supposed to raise `Exit` here so as not to bypass any other click infrastructure that may be wanting to do some cleanup. (also `return` does nothing because `sys.exit()` immediately exits and also click commands aren't really typed to return anything.

There are a couple of other places that use `sys.exit` that I think need to be fixed.

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+
+from astropy import units as units  # type: ignore
+from astropy.io.votable import ucd  # type: ignore
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+logger = logging.getLogger(__name__)
+# logger.setLevel(logging.DEBUG)
+
+
+class BaseObject(ABC, BaseModel):
```
timj, line 36: What's the purpose of the `ABC`  here when there are no abstract methods?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    @model_validator(mode="before")
+    @classmethod
+    def check_units(cls, values: dict[str, Any]) -> dict[str, Any]:
+        """Check that units are valid."""
+        fits_unit = values.get("fits:tunit")
+        ivoa_unit = values.get("ivoa:unit")
+
+        if fits_unit and ivoa_unit:
+            raise ValueError("Column cannot have both FITS and IVOA units")
+        elif fits_unit:
```
timj, line 156: How about instead:

```python
if fits_unit and ivoa_unit:
    raise ValueError("...")
unit = fits_unit or ivoa_unit
```
? I think that is consistent with the elif/elif/else block here.

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+        if unit is not None:
+            try:
+                units.Unit(unit)
+            except ValueError as e:
+                raise ValueError(f"Invalid unit: {e}")
+
+        return values
+
+
+class Constraint(BaseObject, ABC):
```
timj, line 172: BaseObject already inherits from ABC.

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    """A database table constraint."""
+
+    deferrable: bool | None = False
+    """If `True` then this constraint will be declared as deferrable."""
```
timj, line 176: What does `None` mean then?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+
+    initially: str | None = None
+    """Value for ``INITIALLY`` clause, only used if ``deferrable`` is True."""
```
timj, line 179: Should there be a validation to check that initially is None if deferrable is not True? Or is that fine and deferrable can be changed independently? If deferrable is True is `initially` allowed to be `None`?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+
+    annotations: Mapping[str, Any] | None = dict()
```
timj, line 181:
```suggestion
    annotations: Mapping[str, Any] | None = Field(default_factory=dict)
```
?
 How does this ever get a `None` value?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    referenced_columns: list[str] = Field(alias="referencedColumns")
+
+
+class Table(BaseObject):
+    """A database table."""
+
+    columns: list[Column]
+    """The columns in the table."""
+
+    constraints: list[Constraint] | None = []
```
timj, line 236:
```suggestion
    constraints: list[Constraint] | None = Field(default_factory=list)
```
?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    """The constraints on the table."""
+
+    indexes: list[Index] | None = []
```
timj, line 239:
```suggestion
    indexes: list[Index] | None = Field(default_factory=list)
```
Are this and the above ever really None?

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    """The mysql charset to use for the table.
+
+    For now this is a freeform string but it could be constrained to a list of
+    known charsets in the future.
+    """
+
+    @model_validator(mode="before")
+    @classmethod
+    def create_constraints(cls, values: dict[str, Any]) -> dict[str, Any]:
+        """Create constraints from the `constraints` field."""
```
timj, line 265:
```suggestion
        """Create constraints from the ``constraints`` field."""
```

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+            if isinstance(value, BaseModel | list | dict):
+                _extract_ids(value, id_map, processed)
+
+    elif isinstance(obj, list):
+        for item in obj:
+            _extract_ids(item, id_map, processed)
+
+    elif isinstance(obj, dict):
+        for value in obj.values():
+            _extract_ids(value, id_map, processed)
```
timj, line 326: Do you need to add a final `else` block here for unexpected content? mypy isn't going to help you because you declared `obj` as `Any` rather than `BaseModel | list[BaseModel] | dict[str, BaseModel]`.

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+class Schema(BaseObject):
+    """The database schema."""
+
+    version: SchemaVersion | None = None
+    """The version of the schema."""
+
+    tables: list[Table]
+    """The tables in the schema."""
+
+    id_map: dict[str, Any] = Field(dict(), exclude=True)
```
timj, line 338:
```suggestion
    id_map: dict[str, Any] = Field(default_factory=dict, exclude=True)
```

python/felis/datamodel.py
```diff
@@ -0,0 +1,354 @@
...
+    """The current version of the schema."""
+
+    compatible: list[str] | None = None
+    """The compatible versions of the schema."""
+
+    read_compatible: list[str] | None = None
+    """The read compatible versions of the schema."""
+
+
+def _extract_ids(obj: Any, id_map: dict[str, Any], processed: set[int]) -> None:
```
timj, line 302: It seems to me that the `id_map` dict is `dict[str, BaseModel]`

tests/test_datamodel.py
```diff
@@ -0,0 +1,359 @@
...
+class DataModelTestCase(unittest.TestCase):
+    """Test loading of an entire test SDM schema from a YAML file."""
+
+    schema_obj: Schema
+
+    def setUp(self) -> None:
+        """Load data from test file."""
+        with open(TEST_YAML) as test_yaml:
+            data = yaml.safe_load(test_yaml)
+            self.schema_obj = Schema.model_validate(data)
```
timj, line 53: Do you really want the model to be validated in setup? Shouldn't the validation be in the test itself?

timj (approved):
This looks good to me. Some minor comments.

python/felis/datamodel.py
JeremyMcCormick: Removed the ABC parent class.

python/felis/datamodel.py
JeremyMcCormick: I implemented your improved logic.

python/felis/datamodel.py
JeremyMcCormick: I removed the unnecessary ABC parent class.

python/felis/datamodel.py
JeremyMcCormick: I changed the typing to match your suggestion.

tests/test_datamodel.py
JeremyMcCormick: I changed this to validate the model in a test method.

python/felis/datamodel.py
JeremyMcCormick: I removed the `None` in this definition as it was unnecessary.

python/felis/datamodel.py
JeremyMcCormick: I'll leave these without validation for now, as the features related to these variables aren't used in any of the SDM schemas. I'm not totally sure what the intention was in terms of state constraints between them. Probably `initially` can currently be set to a non-None while `deferrable` is `False`, so in that case it would just be ignored.

python/felis/datamodel.py
JeremyMcCormick: I've removed the `None` here as it was unnecessary.

python/felis/datamodel.py
JeremyMcCormick: If there were unexpected content, the validation should fail, which means this method should never be run, as it is designated as "after." 

But I like the typing suggestion and will implement it.

python/felis/datamodel.py
JeremyMcCormick: I have removed the `None` as it is unnecessary.

python/felis/datamodel.py
```diff
@@ -0,0 +1,350 @@
...
+        """Check that columns or expressions are specified, but not both."""
+        if "columns" in values and "expressions" in values:
+            raise ValueError("Defining columns and expressions is not valid")
+        elif "columns" not in values and "expressions" not in values:
+            raise ValueError("Must define columns or expressions")
+        return values
+
+
+class ForeignKeyConstraint(Constraint):
+    """A foreign key constraint on a table."""
```
gpdf, line 218: add "These will be reflected in TAP_SCHEMA keys/key_columns data."

gpdf (changes requested):
Looks good; I am just asking for a few edits to comments.

python/felis/datamodel.py
```diff
@@ -0,0 +1,350 @@
...
+    """The default value of the column."""
+
+    autoincrement: bool | None = None
+    """Whether the column is autoincremented."""
+
+    mysql_datatype: str | None = Field(None, alias="mysql:datatype")
+    """The MySQL datatype of the column."""
+
+    ivoa_ucd: str | None = Field(None, alias="ivoa:ucd")
+    "The IVOA UCD of the column." ""
```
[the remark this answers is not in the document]
JeremyMcCormick: Which quoting do you mean?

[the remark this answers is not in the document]
timj: I think he means that it's just two strings concatenated rather than:

```python
"""The IVOA UCD of the column."""
```

python/felis/datamodel.py
JeremyMcCormick: Did you want this added as a comment on the class?

python/felis/datamodel.py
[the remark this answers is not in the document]
JeremyMcCormick: Fixed

python/felis/datamodel.py
```diff
@@ -0,0 +1,350 @@
...
+    """Whether this is a TAP principal column; can be either 0 or 1.
+
+    This could be a boolean instead of 0 or 1.
+    """
+
+    votable_arraysize: int | Literal["*"] | None = Field(None, alias="votable:arraysize")
+    """The VOTable arraysize of the column."""
+
+    tap_std: int | None = Field(0, alias="tap:std", ge=0, le=1)
+    """Indicates that this is a TAP standard table.
```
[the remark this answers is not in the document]
JeremyMcCormick: Changed

python/felis/datamodel.py
JeremyMcCormick: Added

gpdf (approved):
Good to go

PR #26, opened 2023-12-19
DM-41890: Fix referencing problem with foreign key objects

JeremyMcCormick:
This is a fix for an issue where the `load-tap` command was failing at the command line due to how PyLD was configured. The default setting for object embedding when performing the JSON-LD "framing" was "@last," which embedded a referenced object once, after which its "@id" value was used, instead. The visitor class in Felis was processing the first constraint (foreign key) under the assumption that it would see a column object and not an "@id" reference. But for subsequent calls, it was only seeing the reference string value, which caused a crash.

When loading TAP, Felis is now configured to use the "@always" setting so that the column object data will be immediately accessible when processing constraints. The signature of the `cli._normalize()` method is updated so that the embed setting can be overridden by the caller, and the function argument maintains the default setting from PyLD, which is "@last." The tap module is updated with minor changes, based on the assumption that column objects will be seen during processing now, instead of references. The "@id" value from the column is used as a lookup into the dictionary which contains the TAP column representation. I have verified that the DDL generated by a `load-tap` dry-run is exactly the same between this branch and `main` (using the DP0.2 schema), so these changes do not appear to effect any other functionality.

As a side effect of these changes, including making the tests run successfully, the JSON output generated by various Felis commands such as `normalize` and `merge`, now contains the full column representation (object) in the constraint, rather than a reference "@id." The YAML output of the `modify-tap` command is also effected because it calls the `_normalize` method and needs "@always" to be enabled. Currently, I do not think these changes are a problem, since these particular features within Felis to generate JSON files and modify the TAP table ordering are unused by any production system or workflow. These side effects can be corrected in the future should it become necessary.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -224,7 +224,7 @@ def modify_tap(start_schema_at: int, files: Iterable[io.TextIOBase]) -> None:
         schema_obj["tap:schema_index"] = schema_index
         graph.extend(jsonld.flatten(schema_obj))
     merged = {"@context": DEFAULT_CONTEXT, "@graph": graph}
-    normalized = _normalize(merged)
+    normalized = _normalize(merged, embed="@always")
     _dump(normalized)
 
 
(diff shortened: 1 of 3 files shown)


gpdf (approved):
This looks fine to me.

I do want to eventually test whether the output of modify-tap, with this change, is still usable if input back into the creation of TAP_SCHEMA.

PR #27, opened 2023-12-20
DM-41289: Make Felis fully pip installable

JeremyMcCormick:
Configuration was updated so that felis will be pip installable using:

```    
pip install felis
```
    
The DM developer documentation was used for guidance:
    
https://developer.lsst.io/stack/building-with-pip.html
    
Existing configuration was moved from setup.cfg to pyproject.toml where relevant.

A new job was added to `build.yaml` which publishes the package to PyPI.

A valid token needs to be added as a project secret called `PYPI_UPLOADS` for the publication to work.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
@@ -34,16 +34,49 @@ jobs:
         run: pip install -r requirements.txt
 
       # We have two cores so we can speed up the testing with xdist
-      - name: Install xdist, openfiles and flake8 for pytest
-        run: pip install "flake8<5" pytest-xdist pytest-openfiles pytest-flake8 pytest-cov
+      - name: Install pytest packages
+        run: pip install pytest pytest-xdist pytest-cov
+
+      - name: List installed packages
+        run: pip list -v
 
       - name: Build and install
-        run: pip install -v .
+        run: pip install -v --no-deps -e .
 
       - name: Run tests
-        run: pytest -r a -v -n 3 --open-files --cov=tests --cov=lsst.pex.config --cov-report=xml --cov-report=term
+        run: pytest -r a -v -n 3 --cov=tests --cov=felis --cov-report=xml --cov-report=term --cov-branch
 
       - name: Upload coverage to codecov
-        uses: codecov/codecov-action@v2
+        uses: codecov/codecov-action@v3
         with:
-          file: ./coverage.xml
+          files: ./coverage.xml
+
+  pypi:
+    runs-on: ubuntu-latest
+    needs: [build_and_test]
+    if: startsWith(github.ref, 'refs/tags/')
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Need to clone everything to embed the version.
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade setuptools wheel build
+
+      - name: Build and create distribution
+        run: python -m build --skip-dependency-check
+
+      - name: Upload
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_UPLOADS }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,62 @@
 [build-system]
-requires = ["setuptools", "lsst-versions"]
+requires = ["setuptools", "lsst-versions >= 1.3.0"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "felis"
+description = "A vocabulary for describing catalogs and acting on those descriptions"
+license = {text = "GNU General Public License v3 or later (GPLv3+)"}
+readme = "README.md"
+authors = [
+    {name="Rubin Observatory Data Management", email="dm-admin@lists.lsst.org"},
+]
+classifiers = [
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Scientific/Engineering :: Astronomy"
+]
+keywords = ["lsst"]
+dependencies = [
+    "astropy >= 4",
+    "sqlalchemy >= 1.4",
+    "click >= 7",
+    "pyyaml >= 6",
+    "pyld >= 2",
+    "pydantic >= 2, < 3"
+]
+requires-python = ">=3.11.0"
+dynamic = ["version"]
+
+[project.urls]
+"Homepage" = "https://github.com/lsst/felis"
+
+[project.optional-dependencies]
+test = [
+    "pytest >= 3.2"
+]
+
+[tool.pytest.ini_options]
+
+[tool.setuptools.packages.find]
+where = ["python"]
+
+[tool.setuptools]
+zip-safe = true
+license-files = ["COPYRIGHT", "LICENSE"]
+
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ sqlalchemy >= 1.4
 click
 pyyaml
 pyld
+pydantic >= 2, < 3
```
timj, line 6: I don't think this change should be part of this PR. The pydantic requirements should be added on #25 shouldn't they?

pyproject.toml
```diff
@@ -1,7 +1,55 @@
 [build-system]
-requires = ["setuptools", "lsst-versions"]
+requires = ["setuptools", "lsst-versions >= 1.3.0"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "lsst-felis"
+description = "Felis is a way of describing database catalogs, scientific and otherwise, in a language and DBMS agnostic way."
+license = {text = "BSD 3-Clause License"}
```
timj, line 8: felis is currently GPLv3 I think.

pyproject.toml
```diff
@@ -1,7 +1,55 @@
...
+readme = "README.md"
+authors = [
+    {name="Rubin Observatory Data Management", email="dm-admin@lists.lsst.org"},
+]
+classifiers = [
+    "Intended Audience :: Developers",
```
timj, line 14: Not really sure about this one. Looking at https://pypi.org/classifiers the options aren't great. `Science/Research` might be better but I don't know that "Information Technology" really means.

pyproject.toml
```diff
@@ -1,7 +1,55 @@
...
+    "License :: OSI Approved :: BSD License",
```
timj, line 15: License.

pyproject.toml
```diff
@@ -1,7 +1,55 @@
...
+    "Programming Language :: Python :: 3.12",
+]
+keywords = ["lsst"]
+dependencies = [
+    "astropy",
+    "sqlalchemy >= 1.4",
+    "click",
+    "pyyaml",
+    "pyld",
+    "pydantic >= 2, < 3"
```
timj, line 28: Remove for now?

pyproject.toml
```diff
@@ -1,7 +1,55 @@
...
+]
+
+dynamic = ["version"]
+requires-python = ">=3.10.0"
```
timj, line 32:
```suggestion
requires-python = ">=3.11.0"
```
? Maybe it does work on 3.10 but you don't list it in the classifiers section.

timj (commented):
This looks okay. Some minor comments.

Since this is a draft I will comment rather than approve. It looks like you need to look at the `setup.cfg` file and make sure that you have everything moved across (and then delete everything in it apart of the flake8 section which has to stay. The empty pytest config needs to move to pyproject.toml as well.

requirements.txt
JeremyMcCormick: There was some simultaneous development where I made the same change in both branches. I'll remove this and make the change in the Pydantic branch.

pyproject.toml
JeremyMcCormick: I will use your suggestion of `Science/Research` - it seems like the best fit.

pyproject.toml
JeremyMcCormick: I've changed this to use your suggested version.

pyproject.toml
JeremyMcCormick: I changed this to: 

`"License :: GNU General Public License v3 (GPLv3)"`.

pyproject.toml
JeremyMcCormick: Changed to:

`license = {text = "GNU General Public License v3 (GPLv3)"}`

and also modified the classifiers section.

JeremyMcCormick:
> This looks okay. Some minor comments.
> 
> Since this is a draft I will comment rather than approve. It looks like you need to look at the `setup.cfg` file and make sure that you have everything moved across (and then delete everything in it apart of the flake8 section which has to stay. The empty pytest config needs to move to pyproject.toml as well.

I moved all the configuration from `setup.cfg` to `pyproject.toml`.

The PR should be ready for review now.

JeremyMcCormick:
@timj 

Current commit log looks like this:

```
pick 88845d0 Initial updates for making felis fully pip installable by project name
pick 37c81e9 Make pyproject.toml and requirements.txt deps the same (FIXUP)
pick 74b6e1a Change Python version to 3.10 (FIXUP)                      
pick e1b91c2 Remove Pydantic requirement                                
pick 0aa33d5 Change intended audience to "Science/Research"             
pick 01341b4 Set Python version requirement to 3.11.0 or greater        
pick 13c7ea2 Use GNU v3 license                                         
pick 23021a6 Add empty pytest section                                   
pick 6b5e75e Move all configuration from setup.cfg to pyproject.toml    
pick da003d0 Add some versions to dependencies                          
pick 8321a19 Change build matrix to python 3.11 and 3.12                
pick c62df1d Add Github workflow section to publish package on PyPI    
```

Do you have any preference about how I should make this look in a rebase?

.github/workflows/build.yaml
```diff
@@ -33,17 +33,59 @@ jobs:
       - name: Install dependencies
         run: pip install -r requirements.txt
 
-      # We have two cores so we can speed up the testing with xdist
-      - name: Install xdist, openfiles and flake8 for pytest
-        run: pip install "flake8<5" pytest-xdist pytest-openfiles pytest-flake8 pytest-cov
+      - name: Install pytest packages
+        shell: bash -l {0}
```
timj, line 37: I don't think you need the `shell` line any more here or below.

.github/workflows/build.yaml
```diff
@@ -33,17 +33,59 @@ jobs:
       - name: Install dependencies
         run: pip install -r requirements.txt
 
-      # We have two cores so we can speed up the testing with xdist
```
timj, line 36: This comment can stay I think?

pyproject.toml
```diff
@@ -1,7 +1,63 @@
...
+    "Topic :: Scientific/Engineering :: Astronomy"
+]
+keywords = ["lsst"]
+dependencies = [
+    "astropy >= 4",
+    "sqlalchemy >= 1.4",
+    "click >= 7",
+    "pyyaml >= 6",
+    "pyld >= 2",
+    "pydantic >= 2, < 3"
```
timj, line 29: I thought we were going to remove this from this PR? This should be on #25 I think.

timj (approved):
Looks good to me.

pyproject.toml
gpdf: "This" meaning just the line about `pydantic`, @timj ?

.github/workflows/build.yaml
gpdf: I think that's right.  I don't see a need for it.

gpdf (commented):
I'm not really qualified to review pip-installability.

Tim's comments seem appropriate, so I look forward to a reply on those.

pyproject.toml
timj: Yes. Following up on my comment on the draft where we agreed pydantic should not be mentioned here.

JeremyMcCormick: This will be merged after the Pydantic PR. As `pyproject.toml` doesn't include the dependencies on the Pydantic branch, it makes sense to put it here instead.

.github/workflows/build.yaml
JeremyMcCormick: I will remove the shell lines and then make sure the checks work properly when rerun without them.

pyproject.toml
timj: Ideally you would add it to the setup.cfg file on the other PR and then this PR would move it to the new place. But if you are merging them both today it doesn't really matter.

JeremyMcCormick: > Ideally you would add it to the setup.cfg file on the other PR and then this PR would move it to the new place. But if you are merging them both today it doesn't really matter.

I didn't change `setup.cfg` on DM-41688, though maybe it should have been updated with the new dependency. It should be resolved once this PR is merged, because that file is changed on this PR to only contain flake8 config and no dependencies.

timj: Exactly, it should be on the other and then this should only be moving it, not adding a new dependency at this point. But since they are both merging today it doesn't matter.

JeremyMcCormick:
I am merging without status checks passing because:

- The 3.10 Python build is hanging, which should go away after the Github workflows are merged.

- The codecov is reporting a failure because coverage percentage decreased, but this was due to a misconfiguration in running pytest.

timj:
I fixed the repo configuration to remove python 3.10 and prevent the required checks from being overridden.

PR #31, opened 2024-01-12
DM-42446: Add RSP schema validation

JeremyMcCormick:
RSP schema validation is implemented using a set of extension classes in a new validation module. The validation rules were based on those in the sdm_schemas repository under `tap-schema/validator`.

These new rules check that:

- All columns have a non-empty description
- All tables have a non-empty description, a TAP_SCHEMA table index, and at least one column flagged as TAP_SCHEMA principal
- The schema has a non-empty description and a set of unique (non-duplicated) TAP_SCHEMA table indices

Tests for the new module and its functionality were added under `tests/test_validation.py`.

A schema type can be selected from the command-line now using a syntax such as:

```
felis validate -s RSP ../sdm_schemas/yml/dp02_dc2.yaml
```

I tested the new validation on the above schema and found some expected errors where at least one TAP_SCHEMA principal column needed to be defined and was missing.

I have also added a flag that will make description required for all objects in the schema. It can be enabled as follows:

```
felis validate --require-description ../sdm_schemas/yml/dp02_dc2.yaml
```

This will require descriptions on _every_ object, including the schema itself, tables, columns, and constraints. (RSP validation will only add the extra description requirement for columns.)

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -26,7 +26,7 @@ from typing import Any, Literal, Sequence
 
 from astropy import units as units  # type: ignore
 from astropy.io.votable import ucd  # type: ignore
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 logger = logging.getLogger(__name__)
 # logger.setLevel(logging.DEBUG)
@@ -44,21 +44,28 @@ __all__ = (
     "Schema",
 )
 
+CONFIG = ConfigDict(
+    populate_by_name=True,  # Populate attributes by name.
+    extra="forbid",  # Do not allow extra fields.
+    use_enum_values=True,  # Use enum values instead of names.
+    validate_assignment=True,  # Validate assignments after model is created.
+)
+"""Pydantic model configuration as described in:
+https://docs.pydantic.dev/2.0/api/config/#pydantic.config.ConfigDict
+"""
+
 
 class BaseObject(BaseModel):
     """Base class for all Felis objects."""
 
-    class Config:
-        """Configuration for the `BaseModel` class."""
+    model_config = CONFIG
+    """Pydantic model configuration."""
 
-        populate_by_name = True
-        """Allow fields to be populated by name."""
-
-        extra = "forbid"
-        """Forbid extra fields."""
-
-        use_enum_values = True
-        """Use the values of `Enum` members instead of their names."""
+    class ValidationConfig:
+        """Validation configuration for the `BaseModel` class which is specific
+        to Felis. For now, this is handled with a global flag to avoid holding
+        this state in all of the relevant datamodel objects.
+        """
 
         _require_description = False
         """Flag to require a description for all objects.
@@ -441,9 +448,9 @@ class Schema(BaseObject):
         rather than change the flag directly in `BaseObject`.
         """
         logger.debug(f"Setting description requirement to '{rd}'")
-        BaseObject.Config._require_description = rd
+        BaseObject.ValidationConfig._require_description = rd
 
     @classmethod
     def is_description_required(cls) -> bool:
         """Return whether a description is required for all objects."""
-        return BaseObject.Config._require_description
+        return BaseObject.ValidationConfig._require_description
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -334,7 +334,7 @@ class SchemaVersion(BaseModel):
     """The read compatible versions of the schema."""
 
 
-class SchemaVisitor:
+class SchemaIdVisitor:
     """Visitor to build a Schema object's map of IDs to objects.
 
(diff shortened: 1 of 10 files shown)


JeremyMcCormick:
Marking this a draft for now as there are some mypy errors I need to fix.

tests/test_validation.py
```diff
@@ -0,0 +1,204 @@
...
+
+import unittest
+
+from pydantic import ValidationError
+
+from felis.datamodel import Schema
+from felis.validation import RspColumn, RspSchema, RspTable, get_schema
+
+
+class DataModelTestCase(unittest.TestCase):
```
JeremyMcCormick, line 30: TODO: Rename to RSPSchemaTest

python/felis/datamodel.py
```diff
@@ -305,14 +327,14 @@ class SchemaVersion(BaseModel):
     current: str
     """The current version of the schema."""
 
-    compatible: list[str] | None = None
+    compatible: list[str] = Field(default_factory=list)
     """The compatible versions of the schema."""
 
-    read_compatible: list[str] | None = None
+    read_compatible: list[str] = Field(default_factory=list)
```
gpdf, line 333: How does this change interact with @andy-slac's concerns?

python/felis/validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+from .datamodel import Column, Schema, Table
+
+logger = logging.getLogger(__name__)
+
+
+class RspColumn(Column):
+    """Column for RSP data validation."""
+
+    description: str = Field(..., min_length=1)
```
gpdf, line 35: Can you make this test for non-whitespace?

python/felis/validation.py
```diff
@@ -0,0 +1,101 @@
...
+
+class RspTable(Table):
+    """Table for RSP data validation.
+
+    The list of columns is overridden to use RspColumn instead of Column.
+
+    Tables for the RSP must have a TAP table index and a valid description.
+    """
+
+    description: str = Field(..., min_length=1)
```
gpdf, line 48: Can you make this test for non-whitespace?

python/felis/validation.py
```diff
@@ -0,0 +1,101 @@
...
+    tap_table_index: int = Field(..., alias="tap:table_index")
+    """Redefine the TAP_SCHEMA table index so that it is required."""
+
+    columns: Sequence[RspColumn]
+    """Redefine columns to include RSP validation."""
+
+    @model_validator(mode="after")  # type: ignore
+    @classmethod
+    def check_tap_principal(cls: Any, tbl: "RspTable") -> "RspTable":
+        """Check that at least one column is flagged as the TAP principal."""
```
gpdf, line 60: "is flagged as 'principal' for TAP purposes."  I particularly want to avoid the "the", which seems to read as if it might be normal to have only one.

gpdf (commented):
Some suggestions added

python/felis/datamodel.py
JeremyMcCormick: This change just makes the variable return an empty list by default rather than a `None`, which seems more user friendly. It should not have any effect on the update for making version itself (a separate field) readable as a single string rather than forcing it to be a list.

andy-slac: I think this is OK, we do not have any existing uses of that for now, so we can be free to use empty list instead of `None` as default.

python/felis/validation.py
JeremyMcCormick: Done

python/felis/validation.py
JeremyMcCormick: I changed this to require a non-empty string of minimum length 3 (since length of 1 or 2 would be non-sensical) after whitespace is stripped for RSP columns and tables. I also added some new tests to verify, and it works correctly.

python/felis/validation.py
JeremyMcCormick: See https://github.com/lsst/felis/pull/31#discussion_r1488588606

gpdf (approved):
Looks good, thanks for responses to comments!

JeremyMcCormick:
> Few minor comments, but I see that you merged already, so it is all fine.

These should all be resolved on [this PR](https://github.com/lsst/felis/pull/40).

tests/test_cli.py
```diff
@@ -127,6 +128,28 @@ def test_merge(self) -> None:
...
+        self.assertEqual(result.exit_code, 0)
+
+    def test_validate_default_with_require_description(self) -> None:
+        """Test validate command with description required."""
+        runner = CliRunner()
+        result = runner.invoke(cli, ["validate", "--require-description", TEST_YAML], catch_exceptions=False)
+        # This state needs to be reset for subsequent tests or some will fail.
+        # This is not an issue when running the actual command line tool, which
+        # will execute in its own system process.
+        Schema.require_description(False)
```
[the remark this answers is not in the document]
JeremyMcCormick: I added a try/catch block here just in case.

PR #33, opened 2024-01-18
DM-42540: Check for duplicates when validating schema

JeremyMcCormick:
The Pydantic model was missing a check for duplicate IDs, which can lead to various errors when using other Felis modules such as tap. This will make a list of duplicates as they are found and then raise an error. The entire list will then be printed out after validation.

The error message will look like this:

```
ERROR:felis:1 validation error for Schema
  Value error, Duplicate IDs found in schema:
    #Object.g_deblend_fluxOverlap
    #Object.u_deblend_fluxOverlap
    #Object.y_deblend_fluxOverlap
    #Object.i_deblend_fluxOverlap
    #Object.r_deblend_fluxOverlap
    #Object.z_deblend_fluxOverlap
 [type=value_error, input_value={'name': 'ImsimSchema', '...its:tunit': 'pixel'}]}]}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/value_error
```

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -300,7 +300,10 @@ class SchemaVersion(BaseModel):
 
 
 def _extract_ids(
-    obj: BaseModel | list[BaseModel] | dict[str, BaseModel], id_map: dict[str, BaseModel], processed: set[int]
+    obj: BaseModel | list[BaseModel] | dict[str, BaseModel],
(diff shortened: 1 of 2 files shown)


gpdf:
What's the situation of this PR / branch, @JeremyMcCormick ?  Anything that I should look at?

PR #34, opened 2024-01-19
DM-42544: Add check for duplicate IDs when validating a schema

JeremyMcCormick:
The function for building a map of IDs to objects was rewritten using a visitor pattern, as the old version using recursion was prone to error and not working consistently. Duplicates are added to a set as the schema is visited. The `model_validator` function will throw an error if any duplicates are found, which will include a complete list of all the duplicate IDs. The new implementation of the ID builder and duplicate check was tested by running on several sdm_schemas which are known to contain duplicate IDs (imsim, hsc). These will be fixed in a different ticket branch on sdm_schemas.


gpdf (approved):
Looks good, covers an important missing bit in the current validation.

PR #35, opened 2024-01-19
DM-41289: Change package name to lsst-felis

JeremyMcCormick:
The attempt to reclaim the package name "felis" on PyPI has not been successful, so we will use "lsst-felis" for now.


gpdf (approved):
As agreed

PR #36, opened 2024-01-19
DM-41029: Update Pydantic model for Schema class

andy-slac:
The `version` attribute can accept either string or `SchemaVersion` instance.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -370,7 +370,7 @@ class SchemaVisitor:
 class Schema(BaseObject):
     """The database schema."""
 
-    version: SchemaVersion | None = None
+    version: SchemaVersion | str | None = None
     """The version of the schema."""
 
     tables: list[Table]
--- a/tests/test_datamodel.py
+++ b/tests/test_datamodel.py
@@ -370,6 +370,32 @@ class SchemaVersionTest(unittest.TestCase):
         sv = SchemaVersion(current="1.0.0")
         self.assertEqual(sv.current, "1.0.0", "current should be '1.0.0'")
 
+        # Check that schema version can be specified as a single string or
+        # an object.
+        data = {
+            "name": "schema",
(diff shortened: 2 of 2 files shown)


andy-slac:
@JeremyMcCormick, could you check my one-line update? I want to be able to specify `version` attribute as a single string in YAML (in addition to `SchemaVersion` dict), that should keep things simple for many potential uses.

JeremyMcCormick:
The current syntax would be:

```
version:
  current: "1.2.3"
```

This change just seems to remove the need for the separate key. Is it needed?

Version functionality is not used in the SDM schemas at the moment. I will discuss with @gpdf if we should allow a string here rather than requiring the keys.

andy-slac:
The documentation (that I wrote) explicitly says that string-only version is allowed too. I think this is useful simplification. None of the schema is using versions yet, but I was trying to add version support to `apdb.yaml` (https://github.com/lsst/sdm_schemas/pull/171) and discovered that your new validation code disallows it.

JeremyMcCormick:
> The documentation (that I wrote) explicitly says that string-only version is allowed too. I think this is useful simplification. None of the schema is using versions yet, but I was trying to add version support to `apdb.yaml` ([lsst/sdm_schemas#171](https://github.com/lsst/sdm_schemas/pull/171)) and discovered that your new validation code disallows it.

Okay, it seems reasonable to add this support. I will leave more comments in a review.

JeremyMcCormick (changes requested):
@andy-slac Can you add a test case for this functionality in `tests/test_datamodel.py` ?

You can duplicate line 318 and set a version string here:

```
Schema(name="testSchema", id="#test_sch_id", tables=[test_tbl])
```

Otherwise, this looks reasonable since it was part of the documentation and possible using the original data model.

JeremyMcCormick (approved):
Thank you for adding the test case. This looks good now.

PR #37, opened 2024-01-22
DM-41245: Add a test to confirm that using undefined YAML keys throws an error

JeremyMcCormick:
The error message will be "Extra inputs are not allowed."


gpdf (approved):
Looks good

PR #38, opened 2024-01-31
DM-42680: Add initial implementation of documentation workflow

JeremyMcCormick:
Build the documentation using a compatible version of documenteer and upload for triggers other than PRs. For PRs, the documentation will be built but not uploaded. The README was the same as the documentation site, so it has been replaced by a more minimal text.


JeremyMcCormick:
@gpdf Let's do a re-review of this PR. I replaced the `README.md` with `README.rst` and added some additional sections.

I don't think we should immediately link to the documentation site at the top, because it is already listed in the sidebar. The documentation as it exists is also not a good starting point for users to learn about the tool, as it needs a pretty major overhaul. Maybe we could revisit this in the future once the actual site is in better shape? (I've still linked to it, just in a section that is closer to the bottom of the new README after the summary section.)

gpdf (approved):
OK to proceed

PR #39, opened 2024-02-06
DM-42787: Process constraints after tables are created in SQL visitor

JeremyMcCormick:
When constraints are processed in order, the use of forward references to columns will cause an exception, as they have not been visited and created yet. Instead, the visitor is changed so that the tables are processed first and then these referenced columns will be available when creating constraints.


gpdf (approved):
Looks like the same two-phase approach we needed elsewhere.  Good to go.

PR #40, opened 2024-02-15
DM-42446: Implement minor fixes and improvements to Python syntax

JeremyMcCormick:
This make changes suggested in review comments [here](https://github.com/lsst/felis/pull/31#pullrequestreview-1881543667
). The original PR was closed and merged before the second review occurred, so this is being done on a separate branch. All comments on the original PR should be resolved.

I made an additional change which was adding `from __future__ import annotations` to all of the Python files where it was missing.

I also added a few more simple tests in `test_datamodel.py` that check for various errors on the description field when the `require_description` flag is turned on.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -137,11 +137,18 @@ class CliTestCase(unittest.TestCase):
     def test_validate_default_with_require_description(self) -> None:
         """Test validate command with description required."""
         runner = CliRunner()
-        result = runner.invoke(cli, ["validate", "--require-description", TEST_YAML], catch_exceptions=False)
-        # This state needs to be reset for subsequent tests or some will fail.
-        # This is not an issue when running the actual command line tool, which
-        # will execute in its own system process.
-        Schema.require_description(False)
+        try:
+            # Wrap this in a try/catch in case an exception is thrown.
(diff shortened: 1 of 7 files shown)


tests/test_cli.py
```diff
@@ -137,7 +137,13 @@ def test_validate_default(self) -> None:
...
         """Test validate command with description required."""
         runner = CliRunner()
-        result = runner.invoke(cli, ["validate", "--require-description", TEST_YAML], catch_exceptions=False)
+        try:
+            # Wrap this in a try/catch in case an exception is thrown.
+            result = runner.invoke(
+                cli, ["validate", "--require-description", TEST_YAML], catch_exceptions=False
+            )
+        except Exception as e:
+            print(e)
```
andy-slac, line 146: Hmm, you probably want to rethrow the exception here, so that test fails. ` Schema.require_description(False)` should go into `finally` then.

andy-slac (approved):
Looks great, one minor comment.

tests/test_cli.py
JeremyMcCormick: Good catch. I've made the change as you suggested.

PR #41, opened 2024-02-16
Improve definition of description field

JeremyMcCormick:
Use Annotated along with TypeAlias to get rid of MyPy errors and improve how description strings are handled.


gpdf (approved):
Looks reasonable.  I can't claim I completely understand it, but it definitely seems to improve reusability.

PR #44, opened 2024-02-21
DM-42970: Use a custom function for printing validation errors

JeremyMcCormick:
The new function prints out the error location, message, and input data, omitting some information from the default printout which does not seem that useful to the end user.

TODO

- [ ] Add a test that will read a YAML schema containing many different types of errors and print them out.
- [ ] Include relevant input data for errors resulting from `field_validator` functions, like when a description is too short (Is this possible or would the checks need to be converted to `model_validator` instead?).

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -24,8 +24,10 @@ from __future__ import annotations
 import io
 import json
 import logging
+import os
 import sys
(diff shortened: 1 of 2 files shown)


JeremyMcCormick:
Closing this for now as there are major conflicts due to changes in the `cli` module.

I will reopen later once this is resolved.

PR #45, opened 2024-02-27
DM-43040: Add command line options for setting log level and file

JeremyMcCormick:
Two command line options are added for setting the Felis log level and, optionally, an output file for log messages.

To set these using command line options:

```
felis --loglevel INFO --logfile ./log.txt validate yml/*.yaml
```

And they may also be set using environment variables:

```
export FELIS_LOGLEVEL=INFO
export FELIS_LOGFILE=./log.txt
```

These settings apply to the root logger which will also affect the child loggers. Different settings for each module logger are probably not needed right now.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -45,12 +45,30 @@ from .validation import get_schema
 
 logger = logging.getLogger("felis")
 
+loglevel_choices = ["CRITICAL", "FATAL", "ERROR", "WARNING", "INFO", "DEBUG"]
+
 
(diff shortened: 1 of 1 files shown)


gpdf (approved):
Looks good

timj:
Can you consider using option names that match what we use in butler?

JeremyMcCormick:
> Can you consider using option names that match what we use in butler?

Yes, I can change the CL options to `log-level` and `log-file` to match Butler.

PR #46, opened 2024-02-28
DM-42680: Update documentation workflow triggers based on SQuaRE template

JeremyMcCormick:
Use the triggers from this template suggested by @jonathansick 

https://github.com/lsst/templates/blob/main/project_templates/square_pypi_package/example/.github/workflows/ci.yaml


gpdf (approved):
LGTM

PR #47, opened 2024-03-13
DM-43079: Add metadata module for encapsulating sqlalchemy binding

JeremyMcCormick:
This is a broadly scoped set of changes for adding a `metadata` module to Felis, which replaces the existing `sql` module. The functionality is similar to the old module, but the new ones uses Pydantic for model validation and adds some additional functionality. A number of other dependent and related changes were made as part of this update. [RFC-1003](https://rubinobs.atlassian.net/browse/RFC-1003) triggers this ticket.

The changes include the following:

- Replaced the `create-all` command with `create` in the CLI to perform a similar function of creating database objects from the schema. (The "all" part seemed redundant and unnecessary.)
- Moved some important functions for variant type handling from `sql` into a new `db._variants` module.
- Moved the `InsertDump` class for dumping and printing DDL statements into the `metadata` module from `cli` and made it able to dump DDL directly to a file.
- [DM-43108](https://rubinobs.atlassian.net/browse/DM-43108) - Added the ability to drop and create schemas in the target database using the CLI tool with the `create` command.
- Added a flag to the `create` command for echoing SQL commands when a real (not mock) connection is being used.
- Added CLI arguments to the `create` command for writing SQL statements directly to an output file.
- Added an additional simple test YAML file used in `test_metadata`.
- Modified and added tests for the `metadata` and `cli` modules; the tests for the `sql` module were deleted.

Minor changes that were included in this branch which are not strictly related to the referenced ticket:

- Added configuration for the `pydeps` utility to `pyproject.toml`
- Commented out a line with `value` setting in `tests/data/test.yml` that was causing errors. (The issue causing this error will be resolved on another ticket branch.)

**TODO**

- [x] Add test that loads a schema into a database and compares its built SQA metadata to that reflected from the database to make sure they match.
- [x] Add test that compares information from a schema object with a built SQA metadata object from `MetaDataBuilder`.
- [x] Check what the schema name needs to be when using the `sqlite` driver and modify CLI handling accordingly; currently it is deliberately left blank but might be settable to "main" instead.
- [x] Change construction of MetaData from a schema object to use a builder pattern to avoid polluting interface with build-related functions that are meant to be private

[RFC-1003]: https://rubinobs.atlassian.net/browse/RFC-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DM-43108]: https://rubinobs.atlassian.net/browse/DM-43108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python/felis/cli.py
```diff
@@ -71,26 +71,58 @@ def cli(log_level: str, log_file: str | None) -> None:
...
-        _insert_dump = InsertDump()
-        engine = create_mock_engine(make_url(engine_url), executor=_insert_dump.dump)
-        _insert_dump.dialect = engine.dialect
+        if dry_run:
+            logger.info("Dry run will be executed")
+        dumper = InsertDump(output_file)
+        engine = create_mock_engine(make_url(engine_url), executor=dumper.dump)
+        dumper.dialect = engine.dialect
+        if output_file:
+            logger.info("Writing SQL output to: " + getattr(output_file, "name", ""))
```
timj, line 118:
```suggestion
            logger.info("Writing SQL output to: %s", output_file.name)
```
Although I don't really know why the getattr is needed since `output_file` has already been checked (does `mypy` want you to write `if output_file is not None`?

python/felis/datamodel.py
```diff
@@ -130,7 +133,7 @@ class Column(BaseObject):
     length: int | None = None
     """The length of the column."""
 
-    nullable: bool = True
+    nullable: bool | None = None
```
timj, line 136: Maybe explain what `None` means in this context?

python/felis/datamodel.py
```diff
@@ -462,3 +465,27 @@ def require_description(cls, rd: bool = True) -> None:
     def is_description_required(cls) -> bool:
         """Return whether a description is required for all objects."""
         return cls.ValidationConfig._require_description
+
+    @classmethod
+    def from_yaml_path(cls, yaml_path: str) -> Schema:
+        with open(yaml_path, "r") as yaml_file:
```
timj, line 471: Can't this method call `from_yaml_file` and so avoid the code duplication?

python/felis/datamodel.py
```diff
@@ -462,3 +465,27 @@ def require_description(cls, rd: bool = True) -> None:
...
+            yaml_data = yaml.load(yaml_file, Loader=yaml.SafeLoader)
+        schema = cls.model_validate(yaml_data)
+        return schema
+
+    @classmethod
+    def from_yaml_file(cls, yaml_file: TextIOBase) -> Schema:
```
timj, line 477: I think `file` in this context is confusing since people first think you mean a file path and not a stream.

python/felis/datamodel.py
```diff
@@ -462,3 +465,27 @@ def require_description(cls, rd: bool = True) -> None:
...
+        yaml_data = yaml.load(yaml_file, Loader=yaml.SafeLoader)
+        schema = cls.model_validate(yaml_data)
+        return schema
+
+    @classmethod
+    def from_yaml_gh(
+        cls, schema_name: str, repo: str, branch: str = "main", dir_append: str | None = None
+    ) -> Schema:
+        gh_url = f"https://raw.githubusercontent.com/{repo}/{branch}/{dir_append}/{schema_name}.yaml"
```
timj, line 486: Or change `from_yaml_path` to support `ResourcePath` and then you don't need the rest of this method because that would work for any URIs.

python/felis/db/_variants.py
```diff
@@ -0,0 +1,90 @@
...
+COLUMN_VARIANT_OVERRIDE = {
+    "mysql:datatype": "mysql",
+    "oracle:datatype": "oracle",
+    "postgresql:datatype": "postgresql",
+    "sqlite:datatype": "sqlite",
+}
+
+DIALECT_MODULES = {MYSQL: mysql, ORACLE: oracle, SQLITE: sqlite, POSTGRES: postgresql}
+
+length_regex = re.compile(r"\((.+)\)")
```
timj, line 51:
```suggestion
_length_regex = re.compile(r"\((\d+)\)")
```
This regex could do with an explanation as to what length you are trying to find. It looks like you expect it to be an integer so I'm not sure why it's not `\d+` instead. Underscore since it's a global that you don't want leaking out the file.

timj (commented):
As requested, I've had a quick look but only at the first few files. Can look again another time.

JeremyMcCormick:
> This is a broadly scoped set of changes for adding a `metadata` module to Felis, which replaces the existing `sql` module. The functionality is similar but the new module uses Pydantic for model validation. A number of other dependent and related changes were made as part of this update. [RFC-1003](https://rubinobs.atlassian.net/browse/RFC-1003) has significant overlap with this update, and merging of these changes can be delayed until there is adequate discussion of that proposal.
> 
> 
> 
> These related changes included the following:
> 
> 
> 
> - Replaced the `create-all` command with `create` in the CLI to perform a similar function of creating database objects from the schema.
> 
> - Moved some important functions for variant type handling from `sql` into a new `db._variants` module.
> 
> - Moved the `InsertDump` class for dumping and printing DDL statements into the `metadata` module from `cli` and made it able to dump DDL directly to a file.
> 
> - [DM-43108](https://rubinobs.atlassian.net/browse/DM-43108) - Added the ability to drop and create schemas in the target database using the CLI tool with the `create` command.
> 
> - Added a flag to the `create` command for echoing SQL commands when a real (not mock as in dry run mode) connection is being used.
> 
> - Added CLI arguments to the `create` command for writing SQL statements directly to an output file.
> 
> - Added an additional simple test YAML file used in `test_metadata`.
> 
> - Modified and added tests for the `metadata` and `cli` modules; the tests for the `sql` module were deleted.
> 
> 
> 
> Minor changes that were included in this branch which are not strictly related to the referenced ticket:
> 
> 
> 
> - Added a few utility functions to `datamodel` for loading YAML files into a `Schema` given a file path or handle.
> 
> - Added configuration for the `pydeps` utility to `pyproject.toml`
> 
> - Commented out a line with `value` setting in `tests/data/test.yml` that was causing errors. (The issue causing this error will be resolved on another ticket branch.)
> 
> 
> 
> **TODO**
> 
> 
> 
> - [ ] Add tests to improve coverage for `metadata` module and make sure all types of database objects are tested, including all constraint types and annotations that affect database objects such as `mysql:datatype`.
> 
> - [ ] Verify that the `create` command works correctly from the command line for all database variants (sqlite, mysql, postgresql), both when updating a database and in dry run mode.
> 
> - [ ] Rebase the branch and combine commits (partially from reviewer feedback).
> 
> - [ ] Check what the schema name needs to be when using the `sqlite` driver and modify CLI handling accordingly; currently it is deliberately left blank but might be settable to "main" instead.
> - [ ] Change construction of MetaData from a schema object to use a builder pattern to avoid polluting interface with build-related functions that are meant to be private
> 
> 
> 
> This PR will be marked as a draft until the TODO list is completed.
> 
> 
> 
> [RFC-1003]: https://rubinobs.atlassian.net/browse/RFC-1003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
> 
> [DM-43108]: https://rubinobs.atlassian.net/browse/DM-43108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Okay, thanks. I need to make some fairly major changes to this anyways and add a bunch of tests. I'll let you know when it is ready for a proper review.

python/felis/cli.py
JeremyMcCormick: mypy complains when I try to directly access the `name` attribute:

```
python/felis/cli.py:119: error: "TextIOBase" has no attribute "name"  [attr-defined]
```

timj: You are telling it it is a `TextIOBase` and then requiring that `name` exists. What class are you really expecting it to be?

python/felis/datamodel.py
JeremyMcCormick: > Maybe explain what `None` means in this context?

The existing logic is that if `nullable` is not provided explicitly, it is set to `False` for columns with a numeric type. Otherwise, it is set to `True`. The `None` value is used to signify that no value was provided in the schema itself so this default logic is triggered. So it probably needs to be left alone for now unless we want to change the existing default behavior.

python/felis/datamodel.py
JeremyMcCormick: I'm not sure what you mean by `ResourcePath` here. I can't find a class with this name in the standard Python library.

python/felis/datamodel.py
timj: What I would like is that this is added to the docstring.

python/felis/datamodel.py
JeremyMcCormick: @timj I will rework these utility functions to be less duplicative with better naming so that the purpose of each is clear.

JeremyMcCormick: I'm going to drop these functions from the PR and put them on a separate ticket.

python/felis/cli.py
JeremyMcCormick: > You are telling it it is a `TextIOBase` and then requiring that `name` exists. What class are you really expecting it to be?

I'm not sure what IO class to use here which has a name attribute that the type system can actually see. (It works fine with duck typing but `name` must be in a child class that it can't see.)

python/felis/db/_variants.py
JeremyMcCormick: I inherited this code and copied it from another module I am deleting, but I can look into this. And I will change to have an underscore so it is clear the object is for internal use only.

python/felis/cli.py
timj: Does 

```python
from typing import IO
...
output_file: IO[str]
```
work?

python/felis/db/_variants.py
timj: If it really is numbers it's looking for then `\d+` seems better.

python/felis/cli.py
JeremyMcCormick: I changed this to `typing.IO` which has a visible `name` attribute.

python/felis/db/_variants.py
JeremyMcCormick: I updated the regular expression to look for an integer and also prepended the variable name with an underscore.

python/felis/datamodel.py
JeremyMcCormick: see https://github.com/lsst/felis/pull/47#discussion_r1525442939

python/felis/datamodel.py
JeremyMcCormick: see https://github.com/lsst/felis/pull/47#discussion_r1525442939

python/felis/datamodel.py
JeremyMcCormick: > What I would like is that this is added to the docstring.

I've added a docstring to the field for clarification on handling of the default `None` value.

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+        """
+        columns: list[Column] = []
+        if isinstance(primary_key_columns, str):
+            columns.append(self._objects[primary_key_columns])
+        else:
+            columns.extend([self._objects[column_id] for column_id in primary_key_columns])
+        return PrimaryKeyConstraint(*columns)
+
+    def build_table(self, table_obj: dm.Table) -> None:
+        """Build a SQA table from a `datamodel.Table` object and it to
```
JeremyMcCormick, line 216: fix comment typo

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+        args: dict[str, Any] = {
+            "name": constraint_obj.name if constraint_obj.name else None,
+            "info": constraint_obj.description if constraint_obj.description else None,
+            "deferrable": constraint_obj.deferrable if constraint_obj.deferrable else None,
+            "initially": constraint_obj.initially if constraint_obj.initially else None,
+        }
+        constraint: Constraint
+        constraint_type = constraint_obj.type
+
+        if constraint_type == "ForeignKey":
```
JeremyMcCormick, line 337: improve logic in order to remove type checks

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+        """Initialize the connection wrapper.
+
+        Parameters
+        ----------
+        engine : `Engine` or `MockConnection`
+            The SQLAlchemy engine or mock connection to wrap.
+        """
+        self.engine = engine
+
+    def execute(self, statement: Any) -> ResultProxy:
```
JeremyMcCormick, line 401: statement type could be narrowed

tests/test_metadata.py
```diff
@@ -0,0 +1,193 @@
...
+from felis import datamodel as dm
+from felis.datamodel import Schema
+from felis.metadata import DatabaseContext, MetaDataBuilder, get_datatype_with_variants
+
+TESTDIR = os.path.abspath(os.path.dirname(__file__))
+TEST_YAML = os.path.join(TESTDIR, "data", "sales.yaml")
+
+
+class MetaDataTestCase(unittest.TestCase):
+    """Test creationg of SQLAlchemy `MetaData` from a `Schema`."""
```
JeremyMcCormick, line 46: comment typo

python/felis/metadata.py
JeremyMcCormick: won't fix for now

python/felis/metadata.py
JeremyMcCormick: won't fix for now

python/felis/cli.py
```diff
@@ -71,27 +71,74 @@ def cli(log_level: str, log_file: str | None) -> None:
...
+)
+@click.option("--drop-if-exists", is_flag=True, help="Drop schema if it already exists in the database")
+@click.option("--echo", is_flag=True, help="Echo database commands as they are executed")
+@click.option("--dry-run", is_flag=True, help="Dry run only to print out commands instead of executing")
+@click.option(
+    "--output-file",
+    "-o",
+    type=click.File(mode="w"),
+    help="Write SQL commands to a file instead of executing",
+    required=False,
```
JeremyMcCormick, line 88: The required flag does not need to be set.

python/felis/cli.py
```diff
@@ -71,27 +71,74 @@ def cli(log_level: str, log_file: str | None) -> None:
...
+    """Create database objects from the Felis file."""
+    yaml_data = yaml.safe_load(file)
+    schema = Schema.model_validate(yaml_data)
+    url_obj = make_url(engine_url)
+    if schema_name:
+        logger.info(f"Overriding schema name with: {schema_name}")
+        schema.name = schema_name
+    elif url_obj.drivername == "sqlite":
+        logger.info("Overriding schema name for sqlite with: main")
+        schema.name = "main"
```
JeremyMcCormick, line 110: This string could be a constant as it is used in several other places.

python/felis/db/_variants.py
```diff
@@ -0,0 +1,90 @@
...
+        raise ValueError(f"Type {variant_type_name} not found in dialect {dialect_name}")
+    variant_type = getattr(dialect, variant_type_name)
+    length_params = []
+    if match:
+        length_params.extend([int(i) for i in match.group(1).split(",")])
+    return variant_type(*length_params)
+
+
+def make_variant_dict(column_obj: Column) -> dict[str, TypeEngine[Any]]:
+    """Handle variant overrides, based on logic from `sql.py`.
```
JeremyMcCormick, line 71: This comment references a deleted fild.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+from felis.db._variants import make_variant_dict
+
+from . import datamodel as dm
+from .db import sqltypes
+from .types import FelisType
+
+logger = logging.getLogger(__name__)
+
+
+class InsertDump:
```
JeremyMcCormick, line 60: Class is misnamed as it dumps all statements, not just inserts.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+            The SQA constraint object.
+
+        Raises
+        ------
+        ValueError
+            If the constraint type is not recognized.
+        TypeError
+            If the constraint object is not the expected type.
+        """
+        args: dict[str, Any] = {
```
JeremyMcCormick, line 328: Some or all of these checks seem unnecessary.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        elif constraint_type == "Check" and isinstance(constraint_obj, dm.CheckConstraint):
+            check_obj: dm.CheckConstraint = constraint_obj
+            expression = check_obj.expression
+            constraint = CheckConstraint(expression, **args)
+        elif constraint_type == "Unique" and isinstance(constraint_obj, dm.UniqueConstraint):
+            uniq_obj: dm.UniqueConstraint = constraint_obj
+            columns = [self._objects[column_id] for column_id in uniq_obj.columns]
+            constraint = UniqueConstraint(*columns, **args)
+        else:
+            raise ValueError(f"Unexpected constraint type: {constraint_type}")
```
JeremyMcCormick, line 351: "Unknown" would be clearer.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        """Initialize the connection wrapper.
+
+        Parameters
+        ----------
+        engine : `Engine` or `MockConnection`
+            The SQLAlchemy engine or mock connection to wrap.
+        """
+        self.engine = engine
+
+    def execute(self, statement: Any) -> ResultProxy:
```
JeremyMcCormick, line 392: statement could have a narrower type.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+            The SQLAlchemy metadata object.
+
+        engine : `Engine` or `MockConnection`
+            The SQLAlchemy engine or mock connection object.
+        """
+        self.engine = engine
+        self.metadata = metadata
+        self.connection = ConnectionWrapper(engine)
+
+    def create_if_not_exists(self) -> None:
```
JeremyMcCormick, line 423: Is there a function on SQA MetaData that does this?

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+            elif db_type == "postgresql":
+                logger.info(f"Creating PG schema: {schema_name}")
+                self.connection.execute(sqa_schema.CreateSchema(schema_name, if_not_exists=True))
+            else:
+                raise ValueError("Unsupported database type:" + db_type)
+        except SQLAlchemyError as e:
+            logger.error(f"Error creating schema: {e}")
+            raise
+
+    def drop_if_exists(self) -> None:
```
JeremyMcCormick, line 452: Is there a function on SQA MetaData that does this?

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+    def build_table(self, table_obj: dm.Table) -> None:
+        """Build a SQA `Table` from a `felis.datamodel.Table` and add it to the
+        `MetaData` object.
+
+        Parameters
+        ----------
+        table_obj : `felis.datamodel.Table`
+            The table object to build the SQA table from.
+        """
+        # Process mysql table options.
```
timj, line 224: The docstring says nothing about mysql but this method seems to be using mysql-specific names. Maybe enhance the code comment here saying why postgres doesn't need this tweak.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        columns: list[Column] = []
+        if isinstance(primary_key_columns, str):
+            columns.append(self._objects[primary_key_columns])
+        else:
+            columns.extend([self._objects[column_id] for column_id in primary_key_columns])
+        return PrimaryKeyConstraint(*columns)
+
+    def build_table(self, table_obj: dm.Table) -> None:
+        """Build a SQA `Table` from a `felis.datamodel.Table` and add it to the
```
timj, line 216:
```suggestion
        """Build a SQA `~sqlalchemy.Table` from a `felis.datamodel.Table` and add it to the
```

timj (approved):
Looks okay. I'm not really a SQLAlchemy expert though. If we have too many "if x == 'postgres'" we should consider per-database subclasses.

tests/data/test.yml
```diff
@@ -249,7 +249,7 @@ tables:
         "@id": "#sdqa_Threshold.createdDate"
         datatype: timestamp
         description: Database timestamp when the record is inserted.
-        value: CURRENT_TIMESTAMP
+        # value: CURRENT_TIMESTAMP  ## This causes an error due to quoting.
```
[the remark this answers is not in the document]
JeremyMcCormick: > What's the plan for this line? If it's not going to be fixed delete it. If it is going to be fixed in the future note the ticket number.

I will make a ticket for this and put it into the comment. I do plan on fixing it.

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+                continue
+            new_params = {}
+            for key, value in params.items():
+                if isinstance(value, str):
+                    new_params[key] = f"'{value}'"
+                elif value is None:
+                    new_params[key] = "null"
+                else:
+                    new_params[key] = value
+            print(sql_str % new_params, file=self.file)
```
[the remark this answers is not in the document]
JeremyMcCormick: > Is it documented somewhere where the compiled sql string has `%` formatting in it?

Can you clarify what you want documented here? 

This is code that @andy-slac wrote - I can ask him if there are questions about it.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        apply_schema_to_tables : `bool`, optional
+            Whether to apply the schema name to the tables.
+        """
+        self.schema = schema
+        if not apply_schema_to_metadata:
+            logger.debug("Schema name will not be applied to metadata")
+        if not apply_schema_to_tables:
+            logger.debug("Schema name will not be applied to tables")
+        self.metadata = MetaData(schema=schema.name if apply_schema_to_metadata else None)
+        self._objects: dict[str, Any] = {}
```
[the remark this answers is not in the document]
JeremyMcCormick: > ?

This would need to be a broader type covering any Pydantic object, but you're right to suggest it could be a better type than `Any`.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        ----------
+        schema : `felis.datamodel.Schema`
+            The new schema object to build the SQA metadata from.
+        """
+        self.schema = schema
+        self.metadata = MetaData(schema=self.schema.name)
+        self._objects = {}
+
+    def build(self) -> MetaData:
+        """Build the SQA tables and constraints from the schema."""
```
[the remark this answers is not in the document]
JeremyMcCormick: > SQA means SQLAlchemy?

Yes, I'll change this to "SQLAlchemy" everywhere so it is clearer.

python/felis/metadata.py
JeremyMcCormick: > The docstring says nothing about mysql but this method seems to be using mysql-specific names. Maybe enhance the code comment here saying why postgres doesn't need this tweak.

I will add documentation on this. Postgres does not have similar table options that require special handling.

python/felis/metadata.py
JeremyMcCormick: > A question for @andy-slac I think.

I'll ask - I couldn't find anything in the SQLAlchemy documentation for `MetaData` which does this.

python/felis/metadata.py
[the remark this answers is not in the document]
timj: I think my point relates back to not having any idea what the parameter for the method was (not Any) and then how somebody generates the string such that %-formatting is something that happens. A note in the docstring would be fine if it gave some context and pointed at some other docs.

python/felis/metadata.py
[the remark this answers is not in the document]
timj: Okay. `pydantic.BaseModel` works for that.

python/felis/datamodel.py
JeremyMcCormick: I will leave this for now. There has been some discussion about changing this default to `False`, but I do not want to do that on this ticket.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        associated `Table`.
+        """
+        for table_obj in self.schema.tables:
+            table = self._objects[table_obj.id]
+            for constraint_obj in table_obj.constraints:
+                constraint = self.build_constraint(constraint_obj)
+                table.append_constraint(constraint)
+
+    def build_constraint(self, constraint_obj: dm.Constraint) -> Constraint:
+        """Build a SQA `Constraint` from a `felis.datamodel.Constraint` object.
```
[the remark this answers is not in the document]
JeremyMcCormick: > I think you need to build the docs and check for bad links. Consider running with `package-docs -n build` and it might tell you all the problems.

I'll find and fix these types of errors in the context of another ticket, as the Felis site doesn't even include the Sphinx API documentation right now. So that configuration needs to be done first before `package-docs` even includes it. 

I did fix all the bad typing references you mentioned, as well as a few others that I found.

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def dump(self, sql: Any, *multiparams: Any, **params: Any) -> None:
+        """Dump the SQL statement to a file or stdout.
+
+        Parameters
+        ----------
+        sql : `Any`
+            The SQL statement to dump.
+
+        multiparams : `Any`
```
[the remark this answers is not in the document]
JeremyMcCormick: This is existing code which was copied from a module that is being deleted, and I don't want to change the typing in the context of this ticket. It will be a very ugly type signature, and this is not a user-facing part of Felis.

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+        """Initialize the insert dumper.
+
+        Parameters
+        ----------
+        file : `TextIOBase` or None, optional
+            The file to write the SQL statements to. If None, the statements
+            will be written to stdout.
+        """
+        self.file = file
+        self.dialect: Any | None = None
```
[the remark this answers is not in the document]
JeremyMcCormick: I've changed this to a SQLAlchemy `Dialect` instead.

python/felis/metadata.py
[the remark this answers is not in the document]
JeremyMcCormick: > I think my point relates back to not having any idea what the parameter for the method was (not Any) and then how somebody generates the string such that %-formatting is something that happens. A note in the docstring would be fine if it gave some context and pointed at some other docs.

This just makes sure that when parameters are provided to an executable statement, the resultant SQL is displayed correctly in the dump with the values inserted. I will make a note of this in the function.

python/felis/metadata.py
```diff
@@ -0,0 +1,509 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class InsertDump:
+    """An Insert Dumper for SQL statements which supports writing messages
+    to stdout or a file.
+    """
+
+    def __init__(self, file: IO[str] | None = None) -> None:
```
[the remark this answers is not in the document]
JeremyMcCormick: > Does that work? It might be clearer that way in terms of typing because since then `self.file` has no `None` to deal with. If `self.file` is only ever passed to `print()` it might not matter.

This seems to have the unwanted consequence that when `sys.stdout` is provided explicitly, the CLI test methods will print to the terminal rather than suppressing output. I am going to leave it as is for now.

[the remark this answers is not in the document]
timj: Ok. I guess the test code overrides the default output.

python/felis/metadata.py
andy-slac: MetaData is for table schema only, to create databases/schemas you need to write the code yourself, and it's not super-portable, so the code in this method looks OK.

python/felis/metadata.py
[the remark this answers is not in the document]
andy-slac: This was written before me, I just moved that code to a new location. That code does look a bit unsafe. It may work better if you just say 
```
    compiled = sql.compile(dialect=self.dialect, compile_kwargs={"literal_binds": True})
    print(str(compiled, file=...))
```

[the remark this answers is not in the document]
JeremyMcCormick: > This was written before me, I just moved that code to a new location. That code does look a bit unsafe. It may work better if you just say 
> 
> ```
> 
>     compiled = sql.compile(dialect=self.dialect, compile_kwargs={"literal_binds": True})
> 
>     print(str(compiled, file=...))
> 
> ```

Thank you for the clarification. I'll try your suggested change.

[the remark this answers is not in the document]
JeremyMcCormick: I've implemented the solution from @andy-slac and it seems to work well.

andy-slac:
General comment, for things done before this ticket. Looking at `datamodel.Column`, do I understand right that there is no way to specify datatype override for non-mysql backends? This is a bit troublesome, I already have one branch of `sdm_schemas` where I had to add `postgresql:datatype` for one of the schemas there. I thought that our schema description is supposed to be more flexible and we could provide overrides for any dialect/backend, even ones that we do not know about now (e.g. cassandra)?

andy-slac (approved):
Looks good, small bunch of comments. I'm also worried that with pydantic our datamodel becomes less flexible than what it was.

JeremyMcCormick:
> General comment, for things done before this ticket. Looking at `datamodel.Column`, do I understand right that there is no way to specify datatype override for non-mysql backends? This is a bit troublesome, I already have one branch of `sdm_schemas` where I had to add `postgresql:datatype` for one of the schemas there. I thought that our schema description is supposed to be more flexible and we could provide overrides for any dialect/backend, even ones that we do not know about now (e.g. cassandra)?

Good catch. That can be added easily for Postgres as a field on Column, and I can do it on this ticket.

andy-slac:
> Good catch. That can be added easily for Postgres as a field on Column, and I can do it on this ticket.

I'm OK with this ticket or next, but I would prefer a more general solution, something like:
```
class Column(...):

    dialect_datatypes: dict[str, str]
```
which would be populated from all possible `mysql:datatype`, `postgresql:datatype`, etc. That would probably make it more dynamic than what pydantic likes, but I think this is how original felis was supposed to function?

JeremyMcCormick:
> > Good catch. That can be added easily for Postgres as a field on Column, and I can do it on this ticket.
> 
> I'm OK with this ticket or next, but I would prefer a more general solution, something like:
> 
> ```
> class Column(...):
> 
>     dialect_datatypes: dict[str, str]
>     """Mapping of the dialect names to corresponding datatype overrides."""
> ```
> 
> which would be populated from all possible `mysql:datatype`, `postgresql:datatype`, etc. That would probably make it more dynamic than what pydantic likes, but I think this is how original felis was supposed to function?

I had been thinking about a more generic solution to this as well, but it would be outside the scope of this ticket. I'll keep it in mind going forward.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        args: dict[str, Any] = {
+            "name": constraint_obj.name if constraint_obj.name else None,
+            "info": constraint_obj.description if constraint_obj.description else None,
+            "deferrable": constraint_obj.deferrable if constraint_obj.deferrable else None,
+            "initially": constraint_obj.initially if constraint_obj.initially else None,
+        }
+        constraint: Constraint
+        constraint_type = constraint_obj.type
+
+        if constraint_type == "ForeignKey" and isinstance(constraint_obj, dm.ForeignKeyConstraint):
```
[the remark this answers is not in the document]
JeremyMcCormick: I should be able to drop the redundant check on `constraint_type` here but `@type` has to be kept in the model as it is the only field that distinguishes between different constraint types.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+
+        # Set default value of nullable based on column type and then whether
+        # it was explicitly provided in the schema data.
+        nullable_default = False if isinstance(datatype, Numeric) else True
+        nullable = column_obj.nullable if column_obj.nullable is not None else nullable_default
+
+        # Set autoincrement depending on if it was provided explicitly.
+        autoincrement: Literal["auto", "ignore_fk"] | bool = (
+            column_obj.autoincrement if column_obj.autoincrement else "auto"
+        )
```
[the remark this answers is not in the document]
JeremyMcCormick: Did you mean `if column_obj.autoincrement is not None`?

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        self.schema = schema
+        if not apply_schema_to_metadata:
+            logger.debug("Schema name will not be applied to metadata")
+        if not apply_schema_to_tables:
+            logger.debug("Schema name will not be applied to tables")
+        self.metadata = MetaData(schema=schema.name if apply_schema_to_metadata else None)
+        self._objects: dict[str, Any] = {}
+        self.apply_schema_to_tables = apply_schema_to_tables
+
+    def reset(self, schema: Schema) -> None:
```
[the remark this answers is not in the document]
JeremyMcCormick: I like the idea of removing `reset()` and just using a new builder so I will make that change.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        """Reset the builder with a new schema.
+
+        Parameters
+        ----------
+        schema : `felis.datamodel.Schema`
+            The new schema object to build the SQLAlchemy metadata from.
+        """
+        self.schema = schema
+        self.metadata = MetaData(schema=self.schema.name)
```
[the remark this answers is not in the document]
JeremyMcCormick: See https://github.com/lsst/felis/pull/47#discussion_r1543380244

JeremyMcCormick:
> Looks good, small bunch of comments. I'm also worried that with pydantic our datamodel becomes less flexible than what it was.

We now have to put fields explicitly into the Pydantic data model that we want to access from Python code in Felis. So it is in some sense less flexible than before, but we gain the ability to do strict syntax checking on the schemas. If this turns out to be overly cumbersome, there are ways in Pydantic that we can use "before" validators parse data and then represent it in a more generic form within the created objects (e.g., for variant datatypes).

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+                f"Column '{column_obj.name}' has sized type '{column_obj.datatype}' but no length"
+            )
+        datatype = datatype_fun(column_obj.length, **variant_dict)
+    else:
+        datatype = datatype_fun(**variant_dict)
+    return datatype
+
+
+class MetaDataBuilder:
+    """A class for building a `MetaData` object from a Felis `Schema`."""
```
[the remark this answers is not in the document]
JeremyMcCormick: The idea is that a user could extend the builder class if they so wished.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+        self.build_constraints()
+        return self.metadata
+
+    def build_tables(self) -> None:
+        """Build the SQLAlchemy tables from the schema.
+
+        Notes
+        -----
+        This is the main function for building the SQLAlchemy tables from the
+        schema including objects within tables such as constraints, primary
```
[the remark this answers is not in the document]
JeremyMcCormick: I'll rework this comment. It is somewhat confusing right now.

python/felis/metadata.py
```diff
@@ -0,0 +1,500 @@
...
+            *[self._objects[column_id] for column_id in ensure_iterable(primary_key_columns)]
+        )
+
+    def build_table(self, table_obj: dm.Table) -> None:
+        """Build a `sqlalchemy.Table` from a `felis.datamodel.Table` and add
+        it to the `sqlalchemy.MetaData` object.
+
+        Several MySQL table options are handled by annotations on the table,
+        including the engine and charset. This is not needed for Postgres,
+        which does not have equivalent options.
```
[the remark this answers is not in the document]
JeremyMcCormick: > Postgres has different table options, I guess we do not use any of them presently, but we may want in the future.

I didn't see any Postgres table options in the SQLAlchemy `MetaData` [documentation](https://docs.sqlalchemy.org/en/20/core/metadata.html). But if there are specific ones you think we may use then please point me to the documentation on this.

python/felis/metadata.py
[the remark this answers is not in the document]
JeremyMcCormick: @andy-slac and I will look at this on another ticket branch. There is probably an issue in the TAP visitor code preventing the suggested new code from working correctly.

JeremyMcCormick:
Reduction in test coverage is noted and will be dealt with in subsequent tickets.

PR #48, opened 2024-03-14
DM-41341: Use vounit format when validating units

JeremyMcCormick:
This will apply stricter checks on units than the default. Also set parse_strict to always raise exceptions. This is the default, but make it clear in the validation function that this is the desired behavior when parsing column units.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -195,7 +195,7 @@ class Column(BaseObject):
 
         if unit is not None:
             try:
-                units.Unit(unit)
+                units.Unit(unit, format="vounit", parse_strict="raise")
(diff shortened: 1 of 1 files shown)


JeremyMcCormick:
@gpdf This is marked as a draft for now and should not be merged. The "vounit" unit parser in astropy does not appear to be doing the right thing when it sees completely invalid units. I've filed a [ticket on the astropy github](https://github.com/astropy/astropy/issues/16199) about this.

JeremyMcCormick:
I am going to take this down, because we will have to wait on an astropy update for a fix, which could take awhile.

gpdf:
Good call.

PR #49, opened 2024-03-25
DM-42935: Refactor tap module to use the Pydantic data model

JeremyMcCormick:
The tap module is updated to use the Pydantic data model, instead of the raw YAML. The "load-tap" command in the command line interface was also updated to use the Pydantic schema object. The PYLD transformations were removed from the cli function, as they seemed to be unnecessary. A few changes were made to the Pydantic data model to support setting of reasonable defaults. A "votable_datatype" annotation was added to Colum. The Column datatype was changed to return the enum instead of a string, as this setting seemed to confuse mypy. Tests were changed to conform to the new tap module and the minor changes to the data model. Testing showed that the schema was loaded correctly into a live PostgreSQL database.

- [x] Diff the DDL output for the TAP_SCHEMA from main (production) with this branch.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -393,7 +393,7 @@ class Table(BaseObject):
     primary_key: str | list[str] | None = Field(None, alias="primaryKey")
     """The primary key of the table."""
 
-    tap_table_index: int = Field(0, alias="tap:table_index")
+    tap_table_index: int | None = Field(None, alias="tap:table_index")
     """The IVOA TAP_SCHEMA table index of the table."""
 
     mysql_engine: str | None = Field(None, alias="mysql:engine")
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -93,18 +93,20 @@ class BaseObject(BaseModel):
     description: DescriptionStr | None = None
     """A description of the database object."""
 
-    @model_validator(mode="after")  # type: ignore[arg-type]
-    @classmethod
-    def check_description(cls, object: BaseObject, info: ValidationInfo) -> BaseObject:
+    votable_utype: str | None = Field(None, alias="votable:utype")
+    """The VOTable utype (usage-specific or unique type) of the object."""
+
+    @model_validator(mode="after")
+    def check_description(self, info: ValidationInfo) -> BaseObject:
         """Check that the description is present if required."""
         context = info.context
         if not context or not context.get("require_description", False):
-            return object
-        if object.description is None or object.description == "":
+            return self
+        if self.description is None or self.description == "":
             raise ValueError("Description is required and must be non-empty")
-        if len(object.description) < DESCR_MIN_LENGTH:
+        if len(self.description) < DESCR_MIN_LENGTH:
             raise ValueError(f"Description must be at least {DESCR_MIN_LENGTH} characters long")
-        return object
+        return self
 
 
 class DataType(StrEnum):
@@ -222,11 +224,11 @@ class Column(BaseObject):
     """TAP_SCHEMA indication that this column is defined by an IVOA standard.
     """
 
-    votable_utype: str | None = Field(None, alias="votable:utype")
-    """The VOTable utype (usage-specific or unique type) of the column."""
-
     votable_xtype: str | None = Field(None, alias="votable:xtype")
     """The VOTable xtype (extended type) of the column."""
+
+    votable_datatype: str | None = Field(None, alias="votable:datatype")
+    """The VOTable datatype of the column."""
 
     @field_validator("ivoa_ucd")
     @classmethod
@@ -391,7 +393,7 @@ class Table(BaseObject):
     primary_key: str | list[str] | None = Field(None, alias="primaryKey")
     """The primary key of the table."""
 
-    tap_table_index: int | None = Field(None, alias="tap:table_index")
+    tap_table_index: int = Field(0, alias="tap:table_index")
     """The IVOA TAP_SCHEMA table index of the table."""
 
     mysql_engine: str | None = Field(None, alias="mysql:engine")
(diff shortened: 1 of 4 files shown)


python/felis/tap.py
```diff
@@ -34,14 +34,13 @@
 from sqlalchemy.schema import MetaData
 from sqlalchemy.sql.expression import Insert, insert
 
-from .check import FelisValidator
-from .types import FelisType
-from .visitor import Visitor
+import felis.datamodel as dm
```
andy-slac, line 37: Better to avoid absolute imports if you can, `import .datamodel as dm` should work fine here. I'm not a big fan of very short identifiers, I'd probably use `from . import datamodel`.

python/felis/tap.py
```diff
@@ -133,7 +132,7 @@ class Tap11KeyColumns(Tap11Base):
     )
 
 
-class TapLoadingVisitor(Visitor[None, tuple, Tap11Base, None, tuple, None, None]):
+class TapLoadingVisitor:
```
andy-slac, line 135: Maybe rename it to something else as it is not look like visitor pattern anymore, just a regular class that converts `datamodel` into TAP representation.

python/felis/tap.py
```diff
@@ -202,6 +197,8 @@ def visit_schema(self, schema_obj: _Mapping) -> None:
             session.add_all(keys)
             session.add_all(key_columns)
 
+            logger.debug(f"Committing TAP schema: {schema_obj.name}")
+            logger.debug(f"TAP tables: {len(self.tables)}")
```
andy-slac, line 201: It's better to avoid string formatting when using `logging`, saves few CPU cycles if `debug` is not enabled:

            logger.debug("Committing TAP schema: %s", schema_obj.name)
            logger.debug("TAP tables: %s", len(self.tables))

andy-slac (approved):
Looks great, a couple of minor comments.

python/felis/tap.py
JeremyMcCormick: I think you're right that it would be clearer to use the module name rather than shorten it. I will update this.

python/felis/tap.py
andy-slac: It does the same thing, but not in a "visitor" way. The type of the visitor was generic because we had visitors that converted that schema into different types of objects. Visitor patter was probably not super-useful for that purpose, but this is how it was done initially.

JeremyMcCormick:
@gpdf Following our discussion of validating the new `tap` module which uses Pydantic, I have attached below the existing SQL output from main and that from the updated module. They are exactly the same according to `diff`.

[dp02_dc2_tap_main.sql.txt](https://github.com/lsst/felis/files/14858202/dp02_dc2_tap_main.sql.txt)
[dp02_dc2_tap_new.sql.txt](https://github.com/lsst/felis/files/14858204/dp02_dc2_tap_new.sql.txt)

python/felis/tap.py
JeremyMcCormick: I am just going to leave this for now, as I only removed the complex type signature from the class and did not touch the control flow at all. I didn't change any of its functionality. But I take your point that we may eventually want to name it something else.

JeremyMcCormick:
I've put this into draft mode pending some thorough verification of the changes through testing.

JeremyMcCormick:
@gpdf As per discussion today, I reverted addition of the `tap:schema_index` field on the Pydantic data model, which should have no effect on the TAP_SCHEMA output. The existing scripts within sdm_schemas handle this in a way which does not use the `datamodel` module.

JeremyMcCormick:
@andy-slac I invalidated your review, as there are some fix-ups needed here before merging. I'll have @gpdf review this in its final form.

python/felis/datamodel.py
```diff
@@ -93,18 +93,20 @@ class BaseObject(BaseModel):
     description: DescriptionStr | None = None
     """A description of the database object."""
 
-    @model_validator(mode="after")  # type: ignore[arg-type]
-    @classmethod
-    def check_description(cls, object: BaseObject, info: ValidationInfo) -> BaseObject:
+    votable_utype: str | None = Field(None, alias="votable:utype")
+    """The VOTable utype (usage-specific or unique type) of the object."""
+
+    @model_validator(mode="after")
```
JeremyMcCormick, line 99: These changes are a fix to the Pydantic `model_validator` as it was using the wrong function signature.

python/felis/datamodel.py
```diff
@@ -391,7 +393,7 @@ class Table(BaseObject):
     primary_key: str | list[str] | None = Field(None, alias="primaryKey")
     """The primary key of the table."""
 
-    tap_table_index: int | None = Field(None, alias="tap:table_index")
+    tap_table_index: int = Field(0, alias="tap:table_index")
```
gpdf, line 396: Please retain the ability for the model to distinguish between "missing" and zero `table_index` values.  We may wish to make the handling more sophisticated in the (near) future.

python/felis/tap.py
```diff
@@ -223,56 +225,45 @@ def visit_schema(self, schema_obj: _Mapping) -> None:
...
         table.schema_name = self._schema_name()
-        table.table_name = self._table_name(table_obj["name"])
+        table.table_name = self._table_name(table_obj.name)
         table.table_type = "table"
-        table.utype = table_obj.get("votable:utype")
-        table.description = table_obj.get("description")
-        table.table_index = int(table_obj.get("tap:table_index", 0))
+        table.utype = table_obj.votable_utype
+        table.description = table_obj.description
+        table.table_index = table_obj.tap_table_index
```
gpdf, line 248: This line would have to be changed as well, perhaps to something like 

```python
table.table_index = 0 if table_obj.tap_table_index is None else table_obj.tap_table_index
```

python/felis/datamodel.py
gpdf: I think column_index still allows `None` anyway, right?

gpdf (changes requested):
Looks good in general.  Please make the one change about the handling of missing `table_index` values.

gpdf (approved):
Looks good

PR #50, opened 2024-04-09
DM-41247: Add tool that checks for redundant datatype definitions

JeremyMcCormick:
This adds a Pydantic validator that will identity variant type overrides, e.g., `mysql:datatype` values, that appear to be redundant. To determine whether the datatypes are the same, they are each compiled to a SQL snippet. If the resultant SQL is identical, then the type override is considered to be superfluous and an error is raised.

The validation needs to be turned on using the `-t/--check-redundant-datatypes` flag when running the `validate` command.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -272,8 +272,13 @@ class Column(BaseObject):
 
         datatype_func = get_type_func(datatype)
         felis_type = FelisType.felis_type(datatype)
-        if felis_type.is_sized and length is not None:
-            datatype_obj = datatype_func(length)
+        if felis_type.is_sized:
+            if length is not None:
+                datatype_obj = datatype_func(length)
+            else:
+                raise ValueError(
+                    f"Length must be provided for sized type '{datatype}' in column '{values['@id']}'"
+                )
         else:
             datatype_obj = datatype_func()
 
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -520,8 +520,8 @@ class Schema(BaseObject):
         check_redundant_datatypes = False
         """Flag to enable checking for redundant datatypes on columns.
 
-        An example would be providing both 'mysql:datatype: DOUBLE' and
-        'datatype: DOUBLE' as MySQL would have used that type by default.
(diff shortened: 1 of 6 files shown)


python/felis/datamodel.py
```diff
@@ -404,15 +508,22 @@ def visit_constraint(self, constraint: Constraint) -> None:
...
         """Flag to require a description for all objects.
 
         This is set by the `require_description` class method.
         """
 
+        check_redundant_datatypes = False
+        """Flag to enable checking for redundant datatypes on columns.
+
+        An example would be providing both 'mysql:datatype: DOUBLE' and
+        'datatype: DOUBLE' as MySQL would have used that type by default.
```
JeremyMcCormick, line 524: Double should be lowercase in comment.

tests/test_datatypes.py
```diff
@@ -0,0 +1,128 @@
...
+                "name": self.name,
+                "@id": self.id,
+                "datatype": datatype,
+                f"{self.db_name}:datatype": db_datatype,
+                "length": length,
+            }
+        )
+
+
+def _col_mysql(datatype, mysql_datatype, length=None) -> dict[str, str]:
```
JeremyMcCormick, line 49: Make this generic so it can accept a dialect other than mysql.

tests/test_datatypes.py
```diff
@@ -0,0 +1,128 @@
...
+            with self.assertRaises(ValidationError):
+                # Same type and length
+                _col_mysql("string", "VARCHAR(128)", length=128)
+
+            # Different types, which is okay
+            _col_mysql("double", "FLOAT")
+
+            # Same base type with different lengths, which is okay
+            _col_mysql("string", "VARCHAR(128)", length=32)
+
```
JeremyMcCormick, line 122: Add tests of char vs varchar, unicode, etc.

JeremyMcCormick (commented):
Few changes

gpdf (approved):
Functionally this is what was agreed.  Not enough time to review the code in detail, but it generally looks consistent with the package and our style rules.

PR #51, opened 2024-04-12
DM-43097: Improve support of TEXT type

andy-slac:
Felis insists that `text` datatype is sized and required `length` to be specified. But it maps `text` to MySQL `LONGTEXT` and PostgreSQL `TEXT` types which do not accept length specification. This patch adds a workaround for those two types which drops length argument. A unit test was added as well to trigger various cases for `text` and other sized types

`tests/data/sales.yaml` specified mysql datatype override as `INT` which does not exist (in sqlalchemy at least), it needs to be `INTEGER`.

### DIFF
--- a/python/felis/db/_variants.py
+++ b/python/felis/db/_variants.py
@@ -40,10 +40,10 @@ TABLE_OPTS = {
 }
 
 COLUMN_VARIANT_OVERRIDE = {
-    "mysql:datatype": "mysql",
-    "oracle:datatype": "oracle",
-    "postgresql:datatype": "postgresql",
-    "sqlite:datatype": "sqlite",
+    "mysql_datatype": "mysql",
+    "oracle_datatype": "oracle",
+    "postgresql_datatype": "postgresql",
+    "sqlite_datatype": "sqlite",
 }
 
 DIALECT_MODULES = {MYSQL: mysql, ORACLE: oracle, SQLITE: sqlite, POSTGRES: postgresql}
@@ -87,7 +87,7 @@ def make_variant_dict(column_obj: Column) -> dict[str, TypeEngine[Any]]:
     """
     variant_dict = {}
     for field_name, value in iter(column_obj):
-        if field_name in COLUMN_VARIANT_OVERRIDE:
+        if field_name in COLUMN_VARIANT_OVERRIDE and value is not None:
             dialect = COLUMN_VARIANT_OVERRIDE[field_name]
             variant: TypeEngine = process_variant_override(dialect, value)
             variant_dict[dialect] = variant
--- a/python/felis/db/sqltypes.py
+++ b/python/felis/db/sqltypes.py
@@ -115,10 +115,12 @@ unicode_map: _TypeMap = {
     POSTGRES: postgresql.VARCHAR,
 }
 
-text_map: _TypeMap = {
-    MYSQL: mysql.LONGTEXT,
+# text() method wants to pass length argument to the type, but mysql
+# LONGTEXT and postgres TEXT do not support it.
+text_map: dict[str, Any] = {
+    MYSQL: lambda length: mysql.LONGTEXT(),
     ORACLE: oracle.CLOB,
-    POSTGRES: postgresql.TEXT,
+    POSTGRES: lambda length: postgresql.TEXT(),
 }
 
 binary_map: _TypeMap = {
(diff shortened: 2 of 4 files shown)


JeremyMcCormick (commented):
I am going to reject this, because these changes do not solve the issue with the text type in the manner I had intended. This looks more like a workaround than a proper fix to the type system. I will put up a PR soon with a fix that makes the `text` type in Felis unsized, which I believe is a better way to resolve this issue.

Also, #50 makes some major changes and additions to the facilities for handling database type variants, including a number of new utility functions, which possibly conflict with this PR. We need to merge that PR first and then reassess the database type handling to determine if there are improvements or bug fixes that should be made based on those changes.

python/felis/db/sqltypes.py
```diff
@@ -115,10 +115,12 @@ def compile_double(type_: Any, compiler: Any, **kw: Any) -> str:
     POSTGRES: postgresql.VARCHAR,
 }
 
-text_map: _TypeMap = {
-    MYSQL: mysql.LONGTEXT,
+# text() method wants to pass length argument to the type, but mysql
+# LONGTEXT and postgres TEXT do not support it.
+text_map: dict[str, Any] = {
+    MYSQL: lambda length: mysql.LONGTEXT(),
```
JeremyMcCormick, line 121: I do not like this solution, as it feels more like a workaround than a proper fix for length being required. We should instead make the `text` type itself unsized, since this flag essentially means that length is required, which we do not want for unbounded text fields.

tests/data/sales.yaml
```diff
@@ -70,7 +70,7 @@ tables:
         "@id": "#items.item_id"
         datatype: int
         description: Unique item identifier
-        mysql:datatype: INT
+        mysql:datatype: INTEGER
```
JeremyMcCormick, line 73: I'll remove this override from the test file - it is not needed or currently used in the test assertions.

tests/test_metadata.py
```diff
@@ -188,6 +188,29 @@ def test_builder(self):
                 for primary_key in primary_keys:
                     self.assertTrue(md_table.columns[primary_key].primary_key)
 
+    def test_sized_types(self):
```
JeremyMcCormick, line 191: I like this addition of a test for looking at sized types. Let's include this even if it doesn't not get incorporated via this PR. It can be moved to a user branch instead.

python/felis/db/sqltypes.py
```diff
@@ -203,7 +205,7 @@ def _vary(
     variants.update(overrides)
     for dialect, variant in variants.items():
         # If this is a class and not an instance, instantiate
-        if isinstance(variant, type):
+        if callable(variant):
```
JeremyMcCormick, line 208: This makes sense - we should incorporate this change as well.

JeremyMcCormick (commented):
Some comments on changes unrelated to DM-43097 - some of these should be incorporated, though not necessarily on this branch.

JeremyMcCormick:
Superceded by #52

JeremyMcCormick:
@andy-slac I am going to close this, as it is superseded by #53.

python/felis/db/sqltypes.py
andy-slac: This is a workaround, but it was done according to current documented felis model. If the proper fix requires making `text` type different, it may need an RFC, which means extra delay for me. I will workaround it in dax_ppdb in the meantime.

tests/data/sales.yaml
andy-slac: I made this change because `INT` triggered an error in felis logic after I fixed names `COLUMN_VARIANT_OVERRIDE` (which was totally broken and never tested).

tests/test_metadata.py
andy-slac: I think felis need a lot more unit tests, current test coverage is severely lacking.

python/felis/db/_variants.py
```diff
@@ -40,10 +40,10 @@
 }
 
 COLUMN_VARIANT_OVERRIDE = {
```
JeremyMcCormick, line 42: Need to check if this is a fix for default behavior which is broken.

python/felis/db/sqltypes.py
JeremyMcCormick: > it may need an RFC

I don't believe so since `text` being broken in the past meant no one was using it in their schema files. I believe we can change the `is_sized` flag safely.

PR #52, opened 2024-04-12
DM-43097: Make text type un-sized

JeremyMcCormick:
This changes the Felis `text` type to be un-sized, since `length` should not be required.


JeremyMcCormick:
Wrong ticket - I will put up another PR.

PR #53, opened 2024-04-12
DM-42257: Make the text type un-sized so length is not required

JeremyMcCormick:
The length should not be required when defining a text column, since it represents an unbounded string. The `is_sized` flag in Felis forces a column with that type to require a length, which we do not want for this datatype.

### DIFF
--- a/python/felis/db/_variants.py
+++ b/python/felis/db/_variants.py
@@ -40,10 +40,10 @@ TABLE_OPTS = {
 }
 
 COLUMN_VARIANT_OVERRIDE = {
-    "mysql:datatype": "mysql",
-    "oracle:datatype": "oracle",
-    "postgresql:datatype": "postgresql",
-    "sqlite:datatype": "sqlite",
+    "mysql_datatype": "mysql",
+    "oracle_datatype": "oracle",
+    "postgresql_datatype": "postgresql",
+    "sqlite_datatype": "sqlite",
 }
 
 DIALECT_MODULES = {MYSQL: mysql, ORACLE: oracle, SQLITE: sqlite, POSTGRES: postgresql}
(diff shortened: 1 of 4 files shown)


JeremyMcCormick:
@gpdf proposes changing this line to "char" instead of "unicodeChar" on this PR.

```
class Unicode(FelisType, felis_name="unicode", votable_name="unicodeChar", is_sized=True):
```

JeremyMcCormick:
> @gpdf proposes changing this line to "char" instead of "unicodeChar" on this PR.
> 
> ```
> class Unicode(FelisType, felis_name="unicode", votable_name="unicodeChar", is_sized=True):
> ```

@gpdf I've changed this to "char" on this PR for the Felis text type.

JeremyMcCormick:
@andy-slac I incorporated a number of your unrelated changes (though not all) from #51 in this.

andy-slac (approved):
Looks great! And I mentioned it already - we need unit tests for everything to cover all type-related quirks.

JeremyMcCormick:
> we need unit tests for everything to cover all type-related quirks

I'll keep this in mind and add more on another ticket/branch.

There are a decent number of tests for the type system now but maybe not covering all corner cases.

gpdf:
> > @gpdf proposes changing this line to "char" instead of "unicodeChar" on this PR.
> > ```
> > class Unicode(FelisType, felis_name="unicode", votable_name="unicodeChar", is_sized=True):
> > ```
> 
> @gpdf I've changed this to "char" on this PR for the Felis text type.

The actual change you made was on the `class Text()` line, not on `class Unicode()`, and that's correct, that's exactly what I had in mind: fix `Text()` and leave `Unicode()` alone.

gpdf (approved):
Thanks, looks good.

JeremyMcCormick:
> > > @gpdf proposes changing this line to "char" instead of "unicodeChar" on this PR.
> > > ```
> > > class Unicode(FelisType, felis_name="unicode", votable_name="unicodeChar", is_sized=True):
> > > ```
> > 
> > 
> > @gpdf I've changed this to "char" on this PR for the Felis text type.
> 
> The actual change you made was on the `class Text()` line, not on `class Unicode()`, and that's correct, that's exactly what I had in mind: fix `Text()` and leave `Unicode()` alone.

My mistake on this comment - I actually changed the line with `class Text` and not `class Unicode`.

PR #54, opened 2024-04-12
DM-41247: Revert "Merge pull request #50 from lsst/tickets/DM-41247"

timj:
This reverts commit a32f5f92763923024f883932e3174f911da53f78, reversing changes made to 0d31a4945fe3b5a3dfdc20143ac475f3590b9a53.

Some global state is being left uncleared which is causing test failures with xdist that depend on how many processes are being used.

PR #55, opened 2024-04-15
DM-41247: Implement redundant data type checking with fix for global validation state

JeremyMcCormick:
Revert #54 which itself reverted #50 

Fix the dependence on global state by using [Pydantic context](https://docs.pydantic.dev/latest/concepts/validators/#validation-context) instead.

Summary of changes:

- Use Pydantic `context` object instead of globals for validation flags and remove all class variables from `Schema` controlling validation behavior.
- Turn `validate_assignment` off for now as it can lead to undesirable validation behavior.
- Change the validation of descriptions to an "after" rather than "before" validator and use `ValidationInfo` instead of a global flag.
- Change the `DataType` enum in `datamodel` to `StrEnum` so it can be used interchangeably where a string value is required.
- Add `string_to_typeengine` utility for converting datatype strings to SQLAlchemy types.
- Change validation function for checking redundant datatypes to "after" and use `ValidationInfo`.
- Build the schema's ID map in `model_post_init()` rather than a validator.
- Remove non-standard mapping of `DOUBLE` to `FLOAT` for MySQL and just use SQLAlchemy's default type mapping instead.
- Add new test module `test_datatypes` which checks the validation of when `datatype` and a `mysql:datatype` are both provided in a column description.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -22,6 +22,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from collections.abc import Mapping, Sequence
 from enum import Enum
 from typing import Annotated, Any, Literal, TypeAlias
@@ -29,6 +30,14 @@ from typing import Annotated, Any, Literal, TypeAlias
 from astropy import units as units  # type: ignore
 from astropy.io.votable import ucd  # type: ignore
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+from sqlalchemy import dialects
+from sqlalchemy import types as sqa_types
+from sqlalchemy.engine import create_mock_engine
+from sqlalchemy.engine.interfaces import Dialect
+from sqlalchemy.types import TypeEngine
+
+from .db.sqltypes import get_type_func
+from .types import FelisType
 
 logger = logging.getLogger(__name__)
 
@@ -93,7 +102,7 @@ class BaseObject(BaseModel):
     @classmethod
     def check_description(cls, values: dict[str, Any]) -> dict[str, Any]:
         """Check that the description is present if required."""
-        if Schema.is_description_required():
+        if Schema.Config.require_description:
             if "description" not in values or not values["description"]:
                 raise ValueError("Description is required and must be non-empty")
             if len(values["description"].strip()) < DESCR_MIN_LENGTH:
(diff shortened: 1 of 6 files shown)


python/felis/datamodel.py
```diff
@@ -49,7 +58,7 @@
 CONFIG = ConfigDict(
     populate_by_name=True,  # Populate attributes by name.
     extra="forbid",  # Do not allow extra fields.
-    validate_assignment=True,  # Validate assignments after model is created.
+    validate_assignment=False,  # Validate assignments after model is created.
```
JeremyMcCormick, line 61: This is changed because validating after assignment re-executes validation functions, which is undesirable behavior in some cases. So I'm turning this off for now. It isn't clear that we would ever want a loaded schema to be modified by the user anyways.

python/felis/db/sqltypes.py
```diff
@@ -39,24 +39,12 @@ class TINYINT(SmallInteger):
     __visit_name__ = "TINYINT"
 
 
-class DOUBLE(Float):
```
JeremyMcCormick, line 42: This is not needed anymore.

python/felis/db/sqltypes.py
```diff
@@ -39,24 +39,12 @@ class TINYINT(SmallInteger):
...
-    __visit_name__ = "DOUBLE"
-
-
 @compiles(TINYINT)
 def compile_tinyint(type_: Any, compiler: Any, **kw: Any) -> str:
     """Return type name for TINYINT."""
     return "TINYINT"
 
 
-@compiles(DOUBLE)
```
JeremyMcCormick, line 54: Remove the non-standard double type for MySQL and just use SQLAlchemy's default type mapping.

python/felis/datamodel.py
```diff
@@ -83,40 +91,83 @@ class BaseObject(BaseModel):
...
     description: DescriptionStr | None = None
-    """A description of the database object.
-
-    By default, the description is optional but will be required if
-    `BaseObject.Config.require_description` is set to `True` by the user.
-    """
+    """A description of the database object."""
 
-    @model_validator(mode="before")
+    @model_validator(mode="after")  # type: ignore[arg-type]
```
TallJimbo, line 96: I think I see that you already figured this out on Slack, but if you drop the `@classmethod` I think you won't need the `type: ignore` anymore.

python/felis/datamodel.py
```diff
@@ -83,40 +91,83 @@ class BaseObject(BaseModel):
...
+    int = auto()
+    long = auto()
+    float = auto()
+    double = auto()
+    char = auto()
+    string = auto()
+    unicode = auto()
+    text = auto()
+    binary = auto()
+    timestamp = auto()
```
TallJimbo, line 125: Another option is just
```
DataType: TypeAlias = Literal["boolean", "byte", ...]
```
that will be type-checked by MyPy and validated by Pydantic just as effectively, but it'll otherwise just be plain strings at runtime.

python/felis/datamodel.py
```diff
@@ -83,40 +91,83 @@ class BaseObject(BaseModel):
...
+        raise ValueError(f"Invalid type string: {type_string}")
+
+    type_name, _, params = match.groups()
+    if dialect is None:
+        type_class = getattr(sqa_types, type_name.upper(), None)
+    else:
+        try:
+            dialect_module = _DIALECT_MODULES[dialect.name]
+        except KeyError:
+            raise ValueError(f"Unsupported dialect: {dialect}")
```
TallJimbo, line 155:
```suggestion
            raise ValueError(f"Unsupported dialect: {dialect}") from None
```
or capture the exception instance as `err` and do `from err` instead.  Without one of those you'll get one of those scary-looking, "During handling of exception A, exception B occurred" messages.

python/felis/datamodel.py
```diff
@@ -430,21 +519,29 @@ def check_unique_table_names(cls, tables: list[Table]) -> list[Table]:
...
+    def _create_id_map(self: Schema) -> Schema:
+        """Create a map of IDs to objects.
+
+        This method should not be called by users. It is called automatically
+        by the ``model_post_init()`` method. If the ID map is already
+        populated, this method will return immediately.
+        """
         if len(self.id_map):
-            logger.debug("ID map was already populated")
+            logger.debug("Ignoring call to create_id_map() - ID map was already populated")
```
TallJimbo, line 530: In the line above, `if self.id_map` is sufficient and more idiomatic.

TallJimbo (approved):
Looks good, although this was a somewhat superficial review as I only have a vague sense of what the code is trying to do.  Pydantic usage looks good, though, and I agree this is a good use case for `ValidationInfo.context`.

python/felis/datamodel.py
JeremyMcCormick: I'll leave this for now as I don't want to change downstream code that uses the enum value at this time.

PR #56, opened 2024-04-17
DM-43751: Change default type mapping for boolean to BOOLEAN for MySQL

JeremyMcCormick:
Felis currently maps the boolean type to `BIT(1)` in MySQL, but it would be better to use its standard `BOOLEAN` datatype instead. This would make the usage of `mysql:datatype: BOOLEAN` in some of the columns a redundant override that should be removed, which is intended.

- [x] @gpdf requests to check with @iagaponenko that the live Qserv columns and ingest configuration use the `BOOLEAN` type

### DIFF
--- a/tests/test_datatypes.py
+++ b/tests/test_datatypes.py
@@ -62,9 +62,6 @@ class RedundantDatatypesTest(unittest.TestCase):
             coldata.col("int", "INTEGER")
 
         with self.assertRaises(ValidationError):
-            coldata.col("boolean", "BIT(1)")
-
-        with self.assertRaises(ValidationError):
             coldata.col("float", "FLOAT")
 
         with self.assertRaises(ValidationError):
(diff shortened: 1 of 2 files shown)


tests/test_datatypes.py
```diff
@@ -102,6 +98,9 @@ def test_mysql_datatypes(self) -> None:
             # Same type and length
             coldata.col("string", "VARCHAR(128)", length=128)
```
gpdf, line 99: I'm not entirely sure that I understand the logic that this and the earlier CHAR/VARCHAR/NVARCHAR clauses are testing for objects with lengths.

Is this it?

> A type with a Felis `length` has a redundant override either when the override type matches and is stated without an explicit size (i.e., "VARCHAR" for length-8 "string") or when the override type matches and has an explicit length matching the Felis `length` (i.e. "CHAR(10)" for length-10 "char").

gpdf (approved):
Approved, but if you could comment back on my question about the definition of "redundant" for types with lengths, that would be great.

tests/test_datatypes.py
JeremyMcCormick: > I'm not entirely sure that I understand the logic that this and the earlier CHAR/VARCHAR/NVARCHAR clauses are testing for objects with lengths.
> 
> Is this it?
> 
> > A type with a Felis `length` has a redundant override either when the override type matches and is stated without an explicit size (i.e., "VARCHAR" for length-8 "string") or when the override type matches and has an explicit length matching the Felis `length` (i.e. "CHAR(10)" for length-10 "char").

Yes, a redundant override of this type would look like this in YAML

```
datatype: char
length: 12
mysql:datatype: CHAR(12)
```

These compile to the same SQL of `CHAR(12)` and so they are considered the same.

Does that answer your question?

PR #57, opened 2024-04-19
Improve the log statements and error messages in the redundant type checking validation

JeremyMcCormick:
Overrides that are not considered validation errors will print a `DEBUG` level message like this on one line instead of two now:

```
DEBUG:felis.datamodel:Type override of 'datatype: timestamp' with 'mysql_datatype: DATETIME' in column '#DiaObject.lastNonForcedSource' compiled to 'TIMESTAMP' and 'DATETIME'
```

Those that are redundant and considered validation errors will show like this, with the column's length in the error message (if applicable):

```
Value error, 'mysql_datatype: VARCHAR(64)' is a redundant override of 'datatype: string' in column '#DiaObject.nearbyLowzGal' with length 64
```

Text was changed so that "redundant override" can be used when grep'ing through the logs for errors.


gpdf (approved):
Looks good, thank you!

PR #58, opened 2024-04-22
DM-43753: Make columns nullable by default in the data model

JeremyMcCormick:
The previous approach, which made numeric columns non-nullable while other types were nullable, was not logical considering that SQL columns in DDL are always nullable by default.


andy-slac (approved):
Looks good.

PR #60, opened 2024-04-22
DM-43787: Update Felis documentation workflow

JeremyMcCormick:
Update the Felis documentation workflow to use the latest toolset.

This [example project](https://github.com/lsst/templates/tree/main/project_templates/square_pypi_package/example) was used as a template.


jonathansick (approved):
This looks great 👍 

Down the line, I recommend looking into re-arranging the docs into multiple pages. You can use phalanx.lsst.io, documenteer.lsst.io, or safir.lsst.io as potential inspiration. [We have some documentation on structuring docs in a way that works well with this theme.](https://documenteer.lsst.io/guides/organization.html)

PR #61, opened 2024-04-23
DM-43998: Properly validate default values and fix their handling in SQL metadata

JeremyMcCormick:
This fixes the handling of the `value` field when creating SQLAlchemy metadata. String literals are quoted, except when the value is a SQL function such as `CURRENT_TIMESTAMP`. In this case, quotes are omitted.

A thorough validation function was also added to the data model for checking `value` if it was provided. This will make sure that the provided value is valid given the datatype of the column.

Finally, a new test method was added to check the validation function for all of the Felis datatypes, both when valid and invalid data is provided.

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -37,7 +37,7 @@ from sqlalchemy.engine.interfaces import Dialect
 from sqlalchemy.types import TypeEngine
 
 from .db.sqltypes import get_type_func
-from .types import FelisType
+from .types import Boolean, Byte, Char, Double, FelisType, Float, Int, Long, Short, String, Text, Unicode
 
 logger = logging.getLogger(__name__)
 
@@ -184,7 +184,7 @@ class Column(BaseObject):
     nullable: bool = True
     """Whether the column can be ``NULL``."""
 
-    value: Any = None
+    value: str | int | float | bool | None = None
     """The default value of the column."""
 
     autoincrement: bool | None = None
@@ -224,6 +224,27 @@ class Column(BaseObject):
 
     votable_datatype: str | None = Field(None, alias="votable:datatype")
     """The VOTable datatype of the column."""
+
+    @model_validator(mode="after")
+    def check_value(self) -> Column:
+        """Check that the default value is valid."""
+        if (value := self.value) is not None:
+            if value is not None and self.autoincrement is True:
+                raise ValueError("Column cannot have both a default value and be autoincremented")
+            felis_type = FelisType.felis_type(self.datatype)
+            if felis_type.is_numeric:
+                if felis_type in (Byte, Short, Int, Long) and not isinstance(value, int):
+                    raise ValueError("Default value must be an int for integer type columns")
+                elif felis_type in (Float, Double) and not isinstance(value, float):
+                    raise ValueError("Default value must be a decimal number for float and double columns")
+            elif felis_type in (String, Char, Unicode, Text):
+                if not isinstance(value, str):
+                    raise ValueError("Default value must be a string for string columns")
+                if not len(value):
+                    raise ValueError("Default value must be a non-empty string for string columns")
+            elif felis_type is Boolean and not isinstance(value, bool):
+                raise ValueError("Default value must be a boolean for boolean columns")
+        return self
 
     @field_validator("ivoa_ucd")
     @classmethod
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -276,18 +276,17 @@ class Column(BaseObject):
 
         return values
 
-    @model_validator(mode="after")  # type: ignore[arg-type]
-    @classmethod
-    def validate_datatypes(cls, col: Column, info: ValidationInfo) -> Column:
+    @model_validator(mode="after")
+    def check_datatypes(self, info: ValidationInfo) -> Column:
         """Check for redundant datatypes on columns."""
         context = info.context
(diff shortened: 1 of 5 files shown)


andy-slac (commented):
I am concerned with using JSON formatting for default string value. Do we have a test for `MetadataBuilder` which uses non-empty default for a string column (and generates actual schema)?

python/felis/datamodel.py
```diff
@@ -225,6 +225,31 @@ class Column(BaseObject):
...
+    @model_validator(mode="after")
+    def check_value(self) -> Column:
+        """Check that the default value is valid."""
+        if (value := self.value) is not None:
+            if value is not None and self.autoincrement is True:
+                raise ValueError("Column cannot have both a default value and be autoincremented")
+            felis_type = FelisType.felis_type(self.datatype)
+            if felis_type.is_numeric:
+                if felis_type.felis_name in ("byte", "short", "int", "long"):
+                    if not isinstance(value, int):
```
timj, line 237: If you are worried that a numpy int might turn up then you need to be checking for `numbers.Integral` instead (and then casting to a real `int`).

JeremyMcCormick: > If you are worried that a numpy int might turn up then you need to be checking for `numbers.Integral` instead (and then casting to a real `int`).

These should only ever be `int` or `float` in the `value` field coming from the YAML data. 

This field is not representing the actual column data. It is the setting for the default on the column in the database when generating DDL.

python/felis/datamodel.py
```diff
@@ -225,6 +225,31 @@ class Column(BaseObject):
...
+                        raise ValueError("Default value must be an int for integer type columns")
+                elif felis_type.felis_name in ("float", "double"):
+                    if not isinstance(value, float):
+                        raise ValueError(
+                            "Default value must be a decimal number for float and double columns"
+                        )
```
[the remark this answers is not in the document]
JeremyMcCormick: I agree with this suggestion and have implemented and tested it.

python/felis/metadata.py
```diff
@@ -274,13 +278,19 @@ def build_column(self, column_obj: datamodel.Column) -> Column:
             column_obj.autoincrement if column_obj.autoincrement is not None else "auto"
         )
 
+        # Make sure the default value is properly quoted if it is a string but
+        # ignored if it is a known SQL function.
+        server_default = value
+        if isinstance(server_default, str) and server_default not in _VALID_SERVER_DEFAULTS:
+            server_default = json.dumps(value)
```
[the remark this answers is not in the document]
JeremyMcCormick: You're right that this was not correct and didn't properly handle quoting. I have implemented new logic that seems to work properly now for strings, other types like int, and SQL functions.

All tests have passed and I checked that the default values looked correct in both dry run generated DDL and a live MySQL database.

python/felis/datamodel.py
```diff
@@ -255,51 +280,50 @@ def check_units(cls, values: dict[str, Any]) -> dict[str, Any]:
 
         return values
 
-    @model_validator(mode="after")  # type: ignore[arg-type]
-    @classmethod
-    def validate_datatypes(cls, col: Column, info: ValidationInfo) -> Column:
+    @model_validator(mode="after")
```
JeremyMcCormick, line 283: The changes here are correcting the signature and method body of the validation function. I decided to include this in the PR as a minor fix, though it is unrelated to the ticket.

python/felis/datamodel.py
```diff
@@ -225,6 +225,31 @@ class Column(BaseObject):
...
+            felis_type = FelisType.felis_type(self.datatype)
+            if felis_type.is_numeric:
+                if felis_type in (Byte, Short, Int, Long):
+                    if not isinstance(value, int):
+                        raise ValueError("Default value must be an int for integer type columns")
+                elif felis_type in (Float, Double):
+                    if not isinstance(value, float):
+                        raise ValueError(
+                            "Default value must be a decimal number for float and double columns"
+                        )
```
andy-slac, line 243: Again, maybe add `else` with exception, to be sure that you handle all type in the above two clauses.

andy-slac (approved):
Looks good, one minor comment.

python/felis/metadata.py
```diff
@@ -274,13 +278,21 @@ def build_column(self, column_obj: datamodel.Column) -> Column:
             column_obj.autoincrement if column_obj.autoincrement is not None else "auto"
         )
 
+        server_default: str | TextClause | None = str(value) if value is not None else None
+        if server_default in _VALID_SERVER_DEFAULTS:
+            # If the server default is a valid keyword, use it as is.
+            server_default = text(str(server_default))
```
andy-slac, line 284: Hmm, sorry, I had to re-read this. Does this mean that integer default becomes string and is rendered as string (quoted)?

python/felis/datamodel.py
JeremyMcCormick: I don't think an `else` is necessary here because all of the Felis numeric types are explicitly handled. It should not be possible for `is_numeric` to be `True` and `felis_type` to not be one of these six types.

python/felis/metadata.py
JeremyMcCormick: > Hmm, sorry, I had to re-read this. Does this mean that integer default becomes string and is rendered as string (quoted)?

Yes, this is a bug. I'll fix it so that non-string quotes are used as literals and not quoted.

PR #62, opened 2024-04-24
DM-43597: Remove outdated modules and their tests

JeremyMcCormick:
Remove modules that have been fully deprecated, primarily by the switch to a Pydantic data model rather than using the raw YAML interface.

These modules included:

- `check`: This module has been completely replaced by Pydantic validation.

- `simple`: This is the old data model that was replaced by a Pydantic implementation.

- `visitor`: This is an implementation of a typed visitor pattern, which is not needed anymore.

The corresponding test cases for these modules were removed as well.

The `basic-check` command was removed from the `cli` module and the corresponding test was removed from `test_cli` as well.

- [x] Verify that the Jenkins CI passes on this branch once `dax_apdb` changes are merged to use the Pydantic data model instead of `simple`


JeremyMcCormick:
@andy-slac These changes have been verified on CI in [this build](https://rubin-ci.slac.stanford.edu/job/stack-os-matrix/1242/).

andy-slac (approved):
Looks great!

PR #66, opened 2024-04-26
DM-43371: Add validation of length for sized datatypes

JeremyMcCormick:
Add a standalone validation function that checks if a sized type has a length and raises an error if not.

Add a requirement that the length field in the data model is greater than zero.

Add lengths to some string columns in the tests, as length is always required now for `Column` objects.

This and the related [sdm_schemas PR](https://github.com/lsst/sdm_schemas/pull/212) have been [checked in CI](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/1311/pipeline).

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -178,7 +178,7 @@ class Column(BaseObject):
     datatype: DataType
     """The datatype of the column."""
 
-    length: int | None = None
+    length: int | None = Field(None, gt=0)
     """The length of the column."""
 
     nullable: bool = True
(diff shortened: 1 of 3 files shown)


JeremyMcCormick:
@gpdf Following up on discussion from today, we could add a warning message if a length is provided for a type that is considered un-sized by the Felis type engine. Currently, these would be silently ignored.

gpdf (commented):
> @gpdf Following up on discussion from today, we could add a warning message if a length is provided for a type that is considered un-sized by the Felis type engine. Currently, these would be silently ignored.

I think this sounds like a good idea.

gpdf (approved):
Approved, but with a note that the current "sized type" scheme is too simplistic.

PR #68, opened 2024-04-29
DM-43668: Remove pyld dependency and functionality

JeremyMcCormick:
Remove all functionality in Felis based on the `pyld` library and `JSON-LD` format, as well as the dependence of Felis on this library.

This functionality has been entirely superseded by, and conflicts with, the Pydantic data model.

This includes removal of the following commands from the CLI:

- `modify-tap` - functionality moved to `load-tap` in #49 
- `normalize`
- `merge`
- `dump-json`

The tests for these CLI commands were also removed from `tests/test_cli.py`.

The `util` module and its unit tests were removed, as this included a single visitor-style class, now unused, which depended on `pyld`.

All variables related to `pyld` in `felis/__init__.py` were removed.

Finally, `pyld` itself was removed from `requirements.txt` and `pyproject.toml`.


gpdf (approved):
Without a fully detailed review, this is clearly the way to go.  The JSON-LD approach was not yielding any value in the new model of mapping the Felis specs via Pydantic.

PR #72, opened 2024-06-06
DM-44721: Move database utilities into separate module and refactor them

JeremyMcCormick:
This moves various database functions, classes, and constants from `datamodel` and `metadata` into a new `felis.db.utils` module and implements some minor refactoring. The `InsertDump` class was also renamed to `SQLWriter`.

There should not be any functional differences from these changes.


timj (approved):
Code reorganization is fine.

Note that your codecov reports won't be working -- you will need to switch to action v4 -- see how I do that in daf_butler build.yaml.

Whilst you are there please update the checkout to v4 and the setup-python to v5 in the action yaml files.

JeremyMcCormick:
> Code reorganization is fine.
> 
> Note that your codecov reports won't be working -- you will need to switch to action v4 -- see how I do that in daf_butler build.yaml.
> 
> Whilst you are there please update the checkout to v4 and the setup-python to v5 in the action yaml files.

I updated these github action versions, and codecov is working now. I added the configuration from daf_butler to get the secret.

Thanks!

PR #73, opened 2024-06-07
DM-44059: Add better support for datetime and timestamp

JeremyMcCormick:
Changes:

- Add the datetime type to the Felis type system

- Establish reasonable defaults for datetime and timestamp types for MySQL and PostgreSQL

- Add `precision` and `timezone` column attributes to Pydantic data model

- Update the SQL output in `metadata` to set `precision` and `timezone` from the datamodel fields

- Add sample timestamp and datetime columns to the test schema

- Add tests of SQL output from `timestamp` and `datetime` types

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -125,11 +125,13 @@ class DataType(StrEnum):
     text = auto()
     binary = auto()
     timestamp = auto()
+    datetime = auto()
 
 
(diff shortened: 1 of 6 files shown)


JeremyMcCormick:
This is out of date and in conflict with some big changes from refactoring and reorganizing. I will open another PR on a separate branch.

PR #74, opened 2024-06-10
DM-43012: Move validation checks into the main data model

JeremyMcCormick:
Add new `--check-tap-table-indexes` and `--check-tap-principal` flags to CLI and move the requisite validation methods from the `validation` module into `datamodel`.

Remove the now unneeded `validation` module, fix-up the command line interface, and reorganize/refactor related tests.

Change `Column.check_units` into an "after" `model_validator`.


JeremyMcCormick:
@gpdf I removed the changes to log levels in the CLI. I will change that on a user branch instead.

gpdf (approved):
Looks good.  I like the approach.

PR #76, opened 2024-06-13
DM-44825: Improve support for timestamp handling

JeremyMcCormick:
Implement improvements to timestamps as described in [RFC-1027](https://rubinobs.atlassian.net/browse/RFC-1027):

- Set `timezone=False` when instantiating timestamp types
- Introduce a `precision` attribute on the `Column` data model
- Map the Felis timestamp type to `DATETIME` in MySQL

[RFC-1027]: https://rubinobs.atlassian.net/browse/RFC-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/felis/metadata.py
+++ b/python/felis/metadata.py
@@ -55,16 +55,26 @@ __all__ = ("MetaDataBuilder", "get_datatype_with_variants")
 logger = logging.getLogger(__name__)
 
 
-def _handle_timestamp(
-    column_obj: datamodel.Column, variant_dict: dict[str, TypeEngine[Any]], args: list[Any]
-) -> None:
-    """Handle timezone and precision for timestamp types."""
-    args.append(False)  # Timezone handling is always hard-coded to off.
-    if column_obj.precision:
-        logger.debug("Setting timestamp precision on column '%s' to %d", column_obj.id, column_obj.precision)
-        args.append(column_obj.precision)  # Precision is always the second argument, if provided.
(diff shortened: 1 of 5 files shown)


python/felis/metadata.py
```diff
@@ -54,6 +55,20 @@
...
 
 
+def _handle_timestamp(
+    column_obj: datamodel.Column, variant_dict: dict[str, TypeEngine[Any]], args: list[Any]
+) -> None:
+    """Handle timezone and precision for timestamp types."""
+    args.append(False)  # Timezone handling is always hard-coded to off.
+    if column_obj.precision:
+        logger.debug(
+            "Setting timestamp precision on column '%s' to %d", (column_obj.id, column_obj.precision)
```
timj, line 65:
```suggestion
            "Setting timestamp precision on column '%s' to %d", column_obj.id, column_obj.precision
```

gpdf (approved):
Approved.  Consistent with goal of removing overrides.

Post merge work needed:

1. Clearly explain the timestamp-to-database-entity mappings in the Felis documentation
2. Add validation to Felis to *disallow* the `precision` attribute on all other types, since it's a no-op there and will just be confusing (though we might make a separate decision to use it for a constructive purpose in the future)
3. Make sure the VOTable mapping of `timestamp` is correct, including possibly setting `xtype=timestamp`.

PR #80, opened 2024-06-28
Delete outdated JSON-LD section from user guide

JeremyMcCormick:
The JSON-LD content in the user guide is no longer relevant now that Felis uses a Pydantic data model instead.


gpdf:
What was the ticket under which you removed the actual JSON-LD code?  We can mention it here.

JeremyMcCormick:
> What was the ticket under which you removed the actual JSON-LD code?  We can mention it here.

This was the ticket where I removed it:

https://rubinobs.atlassian.net/browse/DM-43668

PR #82, opened 2024-07-05
DM-44825: Check that precision is only allowed for timestamp columns

JeremyMcCormick:
The `precision` field is used for specifying the precision of the seconds component on timestamp columns and is disallowed for all other datatypes.

This is an extra check on the data model added as part of [DM-44825](https://rubinobs.atlassian.net/browse/DM-44825), which improves timestamp support in Felis.

[DM-44825]: https://rubinobs.atlassian.net/browse/DM-44825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -370,6 +370,19 @@ class Column(BaseObject):
                     )
         return self
 
+    @model_validator(mode="after")
+    def check_precision(self) -> Column:
+        """Check that precision is only valid for timestamp columns.
+
+        Returns
+        -------
+        `Column`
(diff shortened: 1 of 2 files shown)


gpdf (approved):
As noted elsewhere: we may someday enable `precision` for some other data types, but that requires some careful thought, and we don't want people starting to use it before we determine what it's actually supposed to do.

PR #83, opened 2024-07-05
DM-44878: Fill in the TAP_SCHEMA size attribute

JeremyMcCormick:
- Set the `size` attribute based on the value of `votable_arraysize` from the column
    - If arraysize is an int in the YAML data or a string that is parseable as an int, set size=arraysize
    - If arraysize has the form "N*" where N is an int, set size=N
    - For all other cases, size will default to `None` which will result in a null being inserted. This includes the case where arraysize is `*` or a blank string.
- Change the `votable_arraysize` field on the Pydantic model to accept a string or an integer
    - The field was too restrictive based on the valid values from the IVOA VOTable standard.
- Fix bad type hint in `test_tap`

### DIFF
--- a/python/felis/tap.py
+++ b/python/felis/tap.py
@@ -24,6 +24,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from collections.abc import Iterable, MutableMapping
 from typing import Any
 
@@ -125,7 +126,7 @@ def init_tables(
         arraysize = Column(String(10))
         xtype = Column(String(SIMPLE_FIELD_LENGTH))
         # Size is deprecated
-        # size = Column(Integer(), quote=True)
+        size = Column("size", Integer(), quote=True)
         description = Column(String(TEXT_FIELD_LENGTH))
         utype = Column(String(SIMPLE_FIELD_LENGTH))
         unit = Column(String(SIMPLE_FIELD_LENGTH))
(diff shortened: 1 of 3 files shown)


python/felis/tap.py
```diff
@@ -413,6 +414,24 @@ def visit_column(self, column_obj: datamodel.Column, table_obj: Table) -> Tap11B
...
+            except ValueError:
+                return False
+
+        # Handle the deprecated size attribute
+        if arraysize is not None and arraysize != "":
+            if isinstance(arraysize, int):
+                column.size = arraysize
+            elif _is_int(arraysize):
+                column.size = int(arraysize)
+            elif arraysize != "*" and bool(re.match(r"^[0-9]*\*$", arraysize)):
```
gpdf, line 430: Should be `r"^[0-9]+\*$"`

gpdf (commented):
One requested change before merging, see regular expression rewording

python/felis/tap.py
timj: That suggests to me that there was some test coverage missing.

timj: (Which now that I look at the changed files the PR is indeed telling me exactly that...)

JeremyMcCormick: The `tap` module does not have good coverage at all, including for this change. Fixing that is probably beyond the scope of this PR.

gpdf (approved):
Looks good.  Future work on testing is indicated, though.

JeremyMcCormick:
This needs to be fixed up due to changes from #89.

JeremyMcCormick:
This should be fixed now.

PR #85, opened 2024-07-08
Use the pyformat paramstyle for all mock_engines

stvoutsin:
Sets paramstyle=pyformat in the create_mock_engine call in db/utils.py. This addresses the issue where changes to mysql:// as the engine_url breaks at the step of producing the insert statements, as mysql engine in sqlalchemy defaults to named params.
Extends the test_cli.test_load_tap_mock test to run on all known dialects


JeremyMcCormick (approved):
Looks good! Thanks for the fix.

PR #89, opened 2024-07-18
Fix arraysize not being set for text columns

JeremyMcCormick:
Rewrite the logic for setting arraysize so that it is simpler and does not check if the Felis type is sized.

Instead set arraysize from the column or use it's length if arraysize is not set. For timestamp or text columns, set the arraysize to '*' if it was not set explicitly in the schema.

### DIFF
--- a/python/felis/tap.py
+++ b/python/felis/tap.py
@@ -406,12 +406,9 @@ class TapLoadingVisitor:
         felis_type = FelisType.felis_type(felis_datatype.value)
         column.datatype = column_obj.votable_datatype or felis_type.votable_name
 
-        arraysize = None
-        if felis_type.is_sized:
-            arraysize = column_obj.votable_arraysize or column_obj.length or "*"
-        if felis_type.is_timestamp:
-            arraysize = column_obj.votable_arraysize or "*"
-        column.arraysize = arraysize
+        column.arraysize = column_obj.votable_arraysize or column_obj.length
+        if (felis_type.is_timestamp or column_obj.datatype == "text") and column.arraysize is None:
+            column.arraysize = "*"
 
         column.xtype = column_obj.votable_xtype
         column.description = column_obj.description


gpdf (approved):
I think this fixes the immediate issue and can be merged.

However, we are still not correctly generating the "256*" style of arraysize when we do know a maximum length.  So that's another ticket to process in a non-emergency way.

Beyond that, there's also the longer-term issue that we don't properly pass our knowledge of which strings are fixed-length (Felis `char`) and which are variable-length (Felis `string`) through to the VO interface.  But this can't be fixed without verifying that the types of existing character-string columns were really intended to be what they are.

JeremyMcCormick:
> I think this fixes the immediate issue and can be merged.
> 
> However, we are still not correctly generating the "256*" style of arraysize when we do know a maximum length. So that's another ticket to process in a non-emergency way.
> 
> Beyond that, there's also the longer-term issue that we don't properly pass our knowledge of which strings are fixed-length (Felis `char`) and which are variable-length (Felis `string`) through to the VO interface. But this can't be fixed without verifying that the types of existing character-string columns were really intended to be what they are.

Agree with all of this. We should make sure that tickets exist for these issues in Jira.

JeremyMcCormick:
> I think this fixes the immediate issue and can be merged.
> 
> However, we are still not correctly generating the "256*" style of arraysize when we do know a maximum length. So that's another ticket to process in a non-emergency way.
> 
> Beyond that, there's also the longer-term issue that we don't properly pass our knowledge of which strings are fixed-length (Felis `char`) and which are variable-length (Felis `string`) through to the VO interface. But this can't be fixed without verifying that the types of existing character-string columns were really intended to be what they are.

These should be reflected in the Jira now under [DM-45375](https://rubinobs.atlassian.net/browse/DM-45375) and [DM-45376](https://rubinobs.atlassian.net/browse/DM-45376).

PR #90, opened 2024-07-18
DM-45263: Add new tap_schema module

JeremyMcCormick:
This provides the initial implementation for a `tap_schema` module, eventually intended to deprecate and replace the functionality of the Felis `tap` module. It provides the new `load-tap-schema` command in the Felis CLI, which uses the new module to update a TAP_SCHEMA database. The internal representation of TAP_SCHEMA was changed to use a Felis file, found in the `felis.schemas` package.

**Additional Changes**

- Added a few utility functions to `felis.datamodel.Schema` for looking up objects
- Added methods for loading a `Schema` from a file or `lsst.resources.ResourcePath` 
  - Added `lsst-resources` to project requirements
- Added a `test.utils` module with some useful test utility functions
- Updated project configuration to include YAML files from the `felis.schemas` package, which currently includes a standard TAP_SCHEMA representation
- Added a Makefile target for installing project dependencies
- Added the LSST pipelines site to intersphinx dependencies and removed trailing spaces from links (causes problems for some sites)

Relatively complete test coverageis provided for the new module. Some functions marked as not covered actually are tested from loading schema files from `tests/data`. The `tap` module, which this is designed to replace, remains minimally covered in testing.

No breaking changes were intended to be introduced to the API or the CLI by this update. The `felis load-tap` and `init-tap` commands should still work exactly the same as before and can be used side-by-side with the new module for comparison.

----

- [x] Ran in Jenkins CI
- [x] Added a news fragment


python/felis/datamodel.py
```diff
@@ -942,3 +949,114 @@ def __contains__(self, id: str) -> bool:
...
+        context
+            Pydantic context to be used in validation.
+
+        Returns
+        -------
+        `str`
+            The ID of the object.
+        """
+        logger.debug(f"Loading schema from '{resource_path}'")
+        if isinstance(resource_path, str):
```
timj, line 1025: You don't need to do this. You can do `resource_path = ResourcePath(resource_path)` regardless of whether it's a string or ResourcePath. There is a special type you can use for annotating it as well: `"ResourcePathExpression"`.

python/felis/datamodel.py
```diff
@@ -942,3 +949,114 @@ def __contains__(self, id: str) -> bool:
...
+        ValueError
+            If the column is not found in any table.
+        """
+        for table in self.tables:
+            if column in table.columns:
+                return table
+        raise ValueError(f"Column '{column.name}' not found in any table")
+
+    @classmethod
+    def from_resource(cls, resource_path: str | ResourcePath, context: dict[str, Any] = {}) -> Schema:
```
timj, line 1009: By convention we would call this `def from_uri()`. (and annotate with `ResourcePathExpression`.)

python/felis/datamodel.py
```diff
@@ -942,3 +949,114 @@ def __contains__(self, id: str) -> bool:
...
+        context
+            Pydantic context to be used in validation.
+
+        Returns
+        -------
+        `Schema`
+            The schema loaded from the source.
+        """
+        logger.debug("Loading schema from '%s'", source)
+        if isinstance(source, str) or isinstance(source, Path):
```
timj, line 1055: For simplicity you should have this call `from_uri` which can handle `str` and `Path` automatically. So

```python
if isinstance(source, ResourcePathExpression):
    return cls.from_uri(source, context)
...
```

timj (commented):
Just a quick comment on the ResourcePath code.

python/felis/datamodel.py
JeremyMcCormick: What should this look like if it also accepts a `ResourcePath`?

```
    def from_file(cls, source: str | Path | IO[str], context: dict[str, Any] = {}) -> Schema:
```

python/felis/datamodel.py
```diff
@@ -942,3 +986,118 @@ def __contains__(self, id: str) -> bool:
             The ID of the object to check.
         """
         return id in self.id_map
+
+    def find_object_by_id(self, id: str, obj_type: type[T] = BaseObject) -> T:  # type: ignore
```
stvoutsin, line 990: If you don't want to ignore, perhaps casting? (Or not setting a default here) might make things easier: 
```
    def find_object_by_id(self, id: str, obj_type: Type[T] = cast(Type[T], BaseObject)) -> T:
```
But then you also need Schema to be a Generic:

```
class Schema(BaseObject, Generic[T]):
```

python/felis/datamodel.py
```diff
@@ -942,3 +986,118 @@ def __contains__(self, id: str) -> bool:
...
+        logger.debug(f"Loading schema from: '{resource_path}'")
+        try:
+            rp_stream = ResourcePath(resource_path).read()
+        except Exception as e:
+            raise ValueError(f"Error reading resource from '{resource_path}' : {e}") from e
+        yaml_data = yaml.safe_load(rp_stream)
+        return Schema.model_validate(yaml_data, context=context)
+
+    @classmethod
+    def from_stream(cls, source: IO[str], context: dict[str, Any] = {}) -> Schema:
```
stvoutsin, line 1079: It isn't obvious until glancing at the code that this is reading from yaml. Perhaps it would be clearer if the name or the docstring is modified to reflect this?

python/felis/tap_schema.py
```diff
@@ -0,0 +1,633 @@
...
+    def _load_yaml(self) -> None:
+        """Load the standard TAP_SCHEMA schema from a Felis package resource.
+
+        Parameters
+        ----------
+        yaml_file
+            The path to the YAML file to load.
+        """
+        self._load_schema()
+        if self.schema_name != TableManager._SCHEMA_NAME_STD:
```
stvoutsin, line 132:
```suggestion
        if self.schema_name.upper() != TableManager._SCHEMA_NAME_STD:
```

stvoutsin (approved):
Overall I think this looks great. It's pretty clear what all the classes/method are doing, everything is documented and type hinted. I think it will be a useful addition to the cli.

I've added minor nitpicky suggestions, nothing major.

python/felis/tap_schema.py
JeremyMcCormick: Schema names are case sensitive so need to use as is.

python/felis/datamodel.py
JeremyMcCormick: I will update the docstring to note this.

PR #92, opened 2024-07-30
DM-45492: Change names of several create command options

JeremyMcCormick:
Change `--create-if-not-exists` option to `--initialize` and throw an error if the schema already exists in the database. To update an existing database, this option should be omitted. This flag will be ignored for SQLite, as tables can be created in a new database file automatically provided that the engine URL is valid.

Change `--drop-if-exists` option to `--drop`. This will drop and then recreate a schema if it exists. If the schema does not exist, then this is a no-op.

The `--drop` and `--initialize` options were made mutually exclusive and an error will be throw if they are used together. The `--drop` flag always activates initialization, as the database needs to be recreated if it was dropped or does not exist.

Add drop operation support for SQLite, which will remove all the relevant tables from `main` if not in dry run mode.

Add some simple tests of database functionality using `testing.postgresql`. The `TemporaryPostgresInstance` class used for these tests was copied from `daf_butler` and modified, as indicated in the class docstring.

Add a test to `test_cli` which checks that the two new options are mutually exclusive.

Add a Makefile with some convenient commands for local development.

Make a minor change to `cli` so that exceptions in the `create` command are caught and rethrown as a `ClickException`.

Update the Github `build` workflow to install a PostgreSQL testing environment and use `uv` for `pip` commands.

Update the user guide with the new option names.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -165,8 +165,8 @@ def create(
 
         context.create_all()
     except Exception as e:
-        logger.error(e)
-        raise click.exceptions.Exit(1)
+        logger.exception(e)
+        raise click.ClickException(str(e))
 
 
 @cli.command("init-tap", help="Initialize TAP_SCHEMA objects in the database")
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -73,10 +73,10 @@ def cli(log_level: str, log_file: str | None) -> None:
 @cli.command("create", help="Create database objects from the Felis file")
 @click.option("--engine-url", envvar="ENGINE_URL", help="SQLAlchemy Engine URL", default="sqlite://")
 @click.option("--schema-name", help="Alternate schema name to override Felis file")
+@click.option("--initialize", is_flag=True, help="Create the schema in the database if it does not exist")
 @click.option(
-    "--create-if-not-exists", is_flag=True, help="Create the schema in the database if it does not exist"
+    "--drop", is_flag=True, help="Drop schema if it already exists in the database (implies --initialize)"
 )
-@click.option("--drop-if-exists", is_flag=True, help="Drop schema if it already exists in the database")
 @click.option("--echo", is_flag=True, help="Echo database commands as they are executed")
 @click.option("--dry-run", is_flag=True, help="Dry run only to print out commands instead of executing")
 @click.option(
@@ -86,8 +86,8 @@ def cli(log_level: str, log_file: str | None) -> None:
 def create(
     engine_url: str,
     schema_name: str | None,
-    create_if_not_exists: bool,
-    drop_if_exists: bool,
+    initialize: bool,
+    drop: bool,
     echo: bool,
     dry_run: bool,
     output_file: IO[str] | None,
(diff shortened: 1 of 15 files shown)


python/felis/cli.py
```diff
@@ -117,48 +117,56 @@ def create(
     Notes
     -----
     This command creates database objects from the Felis file. The
-    ``--create-if-not-exists`` or ``--drop-if-exists`` flags can be used to
-    create a new MySQL database or PostgreSQL schema if it does not exist
-    already.
+    ``--initialize`` or ``--drop`` flags can be used to create a new MySQL
```
timj, line 120: Not sqlite?

python/felis/cli.py
```diff
@@ -117,48 +117,56 @@ def create(
...
+            context.drop()
+            initialize = True  # If schema is dropped, it needs to be recreated.
+
+        if initialize:
+            logger.debug("Creating schema if not exists")
+            context.initialize()
+
+        context.create_all()
+    except Exception as e:
+        logger.error(e)
```
timj, line 168: `logger.exception`?

python/felis/cli.py
```diff
@@ -117,48 +117,56 @@ def create(
...
+        raise click.ClickException(str(e))
```
timj, line 169: I forget, does click absorb the exception and convert this to an exit status so there is no need for `from e` here?

python/felis/db/utils.py
```diff
@@ -220,13 +220,13 @@ def __init__(self, metadata: MetaData, engine: Engine | MockConnection):
...
 
-    def create_if_not_exists(self) -> None:
+    def initialize(self) -> None:
         """Create the schema in the database if it does not exist.
 
         Raises
         ------
         ValueError
-            If the database is not supported.
+            If the database is not supported or it already exists.
```
timj, line 229: The description above says "create the schema if it does not exist" but this text says it's a ValueError if it already exists. Which is it?

PS General comment that the style guide says a raises block test should start with `Raised if...`

timj (approved):

python/felis/cli.py
JeremyMcCormick: SQLite databases are created automatically by SQLAlchemy if you provide a URL like `sqlite:///example.db` so the option isn't needed - right now it is just silently ignored if provided.

python/felis/cli.py
JeremyMcCormick: That's what it seems to do, yes.

python/felis/db/utils.py
JeremyMcCormick: They shouldn't be mutually exclusive. Does the wording seem unclear?

The `--initialize` flag should be used if you expect the database to not exist. If it already exists, then the flag is not needed. An error will also be raised, because you are expecting to create and update an empty schema/database and don't want to use an existing one.

JeremyMcCormick: > style guide says a raises block test should start with `Raised if...`

I will update this though I have a lot of other `Raises` docstrings that don't follow this style.

python/felis/cli.py
JeremyMcCormick: I'll add some documentation which explains this.

python/felis/db/utils.py
timj: Right. Sorry. It's fine. I was reading too quickly.

JeremyMcCormick: @timj I updated all the `Raises` sections in the docstrings to follow the recommended style.

PR #97, opened 2024-08-05
DM-45485: Add flag for ignoring constraints in schema files

JeremyMcCormick:
Because so many files in SDM Schemas cannot be loaded into a database due to missing indices on foreign key columns, a flag is introduced for ignoring all constraints to avoid these errors, mainly to support testing of database creation for all schemas. Check and unique constraints are currently unused, so no other types of constraints besides foreign key constraints are defined in any of the YAML schema files. Long-term, we may consider adding indices on all FK target columns, but this may not be possible for all schemas.

Add test of this flag for the CLI and when creating metadata.

Also change the environment variable for the database connection URL to `FELIS_ENGINE_URL` from `ENGINE_URL` to avoid possible conflicts with other applications. (I thought this was already what it was. `ENGINE_URL` is too generic.)

Fix some minor type warnings in a unit test.

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -71,7 +71,7 @@ def cli(log_level: str, log_file: str | None) -> None:
 
 
 @cli.command("create", help="Create database objects from the Felis file")
-@click.option("--engine-url", envvar="ENGINE_URL", help="SQLAlchemy Engine URL", default="sqlite://")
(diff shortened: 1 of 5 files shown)


gpdf (approved):
LGTM, discussed in standup today

JeremyMcCormick:
Tested in CI:

https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/2088/pipeline

PR #98, opened 2024-08-12
DM-45623: Add check that constraint names are unique within a schema

JeremyMcCormick:
Add a check at the schema level that constraints are uniquely named.

This is necessary because TAP_SCHEMA uses the `name` field of a foreign key constraint as the primary key of the `keys` table.


gpdf (approved):
LGTM, definitely needed.

The unique-key constraint here is inherent to the TAP_SCHEMA data model.

PR #99, opened 2024-08-12
DM-45725: Change default table engine to MyISAM

JeremyMcCormick:
Change the default MySQL table engine to MyISAM in the data model, for two main reasons:

- This is the table engine used by Qserv, so we want to match it as closely as possible.
- Some of the tables are exceeding the maximum number of columns allowed by the InnoDB engine.


gpdf (approved):
LGTM, matching Qserv is reasonable

PR #100, opened 2024-08-23
DM-45938: Add support for generating IDs on data model objects

JeremyMcCormick:
Add an `--id-generation` flag which turns on generation of `@id` values for schema objects that do not have them.

Add a Schema validator that checks if index names are unique.


docs/changes/DM-45938.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added automatic ID generation for objects in Felis schemas when the `--id-generation` flag is included on the command line.
+This is supported for the `create` and `validate` commands.
```
gpdf, line 2: Please add here that the change also "add[s] validator that checks if index names are unique within the schema"

python/felis/datamodel.py
```diff
@@ -798,6 +852,36 @@ def check_unique_constraint_names(self: Schema) -> Schema:
 
         return self
 
+    @model_validator(mode="after")
+    def check_unique_index_names(self: Schema) -> Schema:
```
gpdf, line 856: As noted above, please add this to the news snippet.

gpdf (changes requested):

gpdf (approved):
Looks good now.  Very much agree with the goal of removing the requirement to tediously supply IDs in the input schema files.

PR #102, opened 2024-08-28
DM-46002: Locations for validation errors on constraints reported incorrectly

JeremyMcCormick:
This fixes an issue where the Pydantic error locations for constraint objects were reported incorrectly, due to how the model objects were constructed.

Instead of using a custom `create_constraints()` function, a Pydantic [discriminated union](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators) was used to automatically construct the objects based on the `type` field. This seems to fix the issue where error locations were reported incorrectly.

Sample new error location is: `tables.0.constraints.0.Unique.@id`.

This is not exactly the same scheme that shows for some other types of objects, but it is at least clear now that the error has occurred in a specific constraint denoted by the index.

As a result of the above changes, usage of the `type` field in the `metadata` and `tap` modules was removed in favor of checking the constraint's type using `isinstance()`. This is required based on removal of the `type` field from the `Constraint` class. (Since it should never be instantiated, the `type` field is not needed on this class and including it creates typing conflicts with its child classes. Each constraint class now defines this separately and not as a common field from the parent class.)

A few additional changes were included which do not have associated tickets:

- Cleaned up the `Constraint` test case in `test_datamodel` (See https://github.com/lsst/felis/pull/102/commits/1706788efc2dda8f089c7d777d4fe7e5d9cb0935).
- Added check to the data model that the `deferrable` field on `Constraint` must be set to `True` if `initially` is not None. This requirement was indicated in comments but never implemented.
- Added a `Literal` to `Constraint` in the data model that limits the value of `initially` to `IMMEDIATE` or `DEFERRED` as these are the only valid values for an `INITIALLY` clause in DDL.
- Removed the unused `annotations` field on the `Constraint` class. (This field was erroneously carried over from the old "simple model" but has never been used within a production schema, to my knowledge.)
- Cleaned up the `Index` test case in `test_datamodel` (See https://github.com/lsst/felis/pull/102/commits/6631610d8da3dacf7b7952df773cfc5b43df34e2).

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/docs/changes/DM-46002.bugfix.rst
+++ b/docs/changes/DM-46002.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where the error locations on constraint objects during validation were reported incorrectly.
+This was accomplished by replacing the ``create_constraints()`` function with a Pydantic `discriminated union <https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators>`__.
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -394,6 +394,20 @@ class Constraint(BaseObject):
     """Value for ``INITIALLY`` clause; only used if `deferrable` is
     `True`."""
 
+    @model_validator(mode="after")
+    def check_deferrable(self) -> Constraint:
+        """Check that the ``INITIALLY`` clause is only used if `deferrable` is
+        `True`.
+
+        Returns
+        -------
+        `Constraint`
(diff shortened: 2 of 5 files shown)


gpdf (approved):
I have to admit to not being qualified to review the Pydantic details involved here; the goals seem appropriate and I'm willing to accept that the resulting messages are improved in their intelligibility.

PR #106, opened 2024-09-26
Fix bugs in how size and arraysize were being handled

JeremyMcCormick:
Fix obvious bug where wrong arraysize variable was being used in setting size.

Do not set arraysize automatically for fields with length of 1.

## Checklist

- [ ] Ran Jenkins

### DIFF
--- a/python/felis/tap.py
+++ b/python/felis/tap.py
@@ -407,7 +407,9 @@ class TapLoadingVisitor:
         felis_type = FelisType.felis_type(felis_datatype.value)
         column.datatype = column_obj.votable_datatype or felis_type.votable_name
 
-        column.arraysize = column_obj.votable_arraysize or column_obj.length
+        column.arraysize = column_obj.votable_arraysize or (
+            column_obj.length if (column_obj.length is not None and column_obj.length > 1) else None
+        )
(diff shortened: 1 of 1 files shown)


python/felis/tap.py
```diff
@@ -407,7 +407,9 @@ def visit_column(self, column_obj: datamodel.Column, table_obj: Table) -> Tap11B
         felis_type = FelisType.felis_type(felis_datatype.value)
         column.datatype = column_obj.votable_datatype or felis_type.votable_name
 
-        column.arraysize = column_obj.votable_arraysize or column_obj.length
+        column.arraysize = column_obj.votable_arraysize or (
+            column_obj.length if (column_obj.length is not None and column_obj.length > 1) else None
```
stvoutsin, line 411: Is there a case where `column_obj.votable_arraysize` is None `column_obj_length == 1` and we want `column.arraysize `set to that?

stvoutsin (approved):
Looks good to me.
I've tested on dev and the size validation errors are all gone. One small comment/question regarding arraysize=1

python/felis/tap.py
JeremyMcCormick: It says in the TAP 1.1 spec that arraysize should never be set to 1.

PR #109, opened 2024-10-30
DM-46130: Add tools for comparing two schemas or a schema and a database

JeremyMcCormick:
This adds a `diff` command to the CLI which performs comparisons of two schemas for differences or a schema and a database via an engine URL.

## Checklist

- [x] Ran Jenkins


.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.11
+        language_version: python3.12
```
timj, line 18: This reminds me that you should now be able to add 3.13 to your build matrix.

.github/workflows/build.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.11", "3.12", "3.13"]
```
timj, line 16: Don't forget to also add 3.13 to the pyproject.toml file list of supported versions (since this did build fine)

pyproject.toml
```diff
@@ -21,15 +21,16 @@ classifiers = [
 ]
 keywords = ["lsst"]
 dependencies = [
-    "astropy >= 4",
-    "sqlalchemy >= 1.4",
-    "click >= 7",
-    "pyyaml >= 6",
-    "pydantic >= 2, < 3",
+    "astropy >= 6.0.0",
+    "numpy < 2",
```
timj: You don't need this if numpy 2 works fine.

pyproject.toml
```diff
@@ -21,15 +21,16 @@ classifiers = [
...
+    "sqlalchemy >= >=2.0.23",
```
timj: This seems wrong.

pyproject.toml
JeremyMcCormick: Using numpy 2 results in an annoying error message printing at the top when the command line tools are run, which is why I added this. The stack is still using a numpy 1.x version as well. (numpy is not a direct dependency of felis though.)

timj: Stack using numpy 1.x is orthogonal to whether we allow other users to install this at all with numpy 2. What's the error message? Is it from astropy?

pyproject.toml
JeremyMcCormick: Which dependency seems wrong here?

pyproject.toml
JeremyMcCormick: No, it's an error message from numpy itself that I had trouble making go away.

timj: What is the error message

.github/workflows/build.yaml
JeremyMcCormick: Done

pyproject.toml
JeremyMcCormick: I would also think that if we'll be doing additional felis development using astropy soon (e.g., generating astropy tables), the numpy version should match that from the stack.

pyproject.toml
JeremyMcCormick: Oh, I see. The `sqlalchemy` version needs to be updated to `>=2.0.23`. Will do...

pyproject.toml
timj: I disagree for the pypi version. It's the whole point of having tests on GitHub rather than locking ourselves to rubinenv (which can easily test for using Jenkins). What is the numpy2 error message?

JeremyMcCormick: numpy error message:

https://github.com/lsst/felis/pull/109#discussion_r1829788091

timj: Thanks. `numpy.longdouble` does exist in numpy 2. It's a bit of an odd message. Really you want the stack trace from that warning.

JeremyMcCormick: I'm going to remove the numpy requirement entirely, given that astropy has it as a transitive dependency and felis does not use it directly.

JeremyMcCormick:
By request of @gpdf, dropping https://github.com/lsst/felis/pull/109/commits/6c234d5d2240f3ec9735026a9c0c87fe29c969ce which adds Python 3.13 to the build matrix.

gpdf (changes requested):
Most requested changes were discussed in our co-working session today.

Summary:

- [ ] Remove Python version change to a separate PR
- [ ] Remove "metadata" file changes to a separate PR
- [ ] Add examples to command-line help
- [ ] Add an option to produce a non-zero return code if there are any differences found (this will be used to do the work in DM-46158)

JeremyMcCormick:
> * [ ]  Remove "metadata" file changes to a separate PR

This is a pain to drop from this PR due to conflicts, as it modifies more than just the `metadata` module and its tests. I'd suggest we just leave it in.

JeremyMcCormick:
@gpdf This is ready for a re-review. All the requested changes from https://github.com/lsst/felis/pull/109#pullrequestreview-2474252367 were implemented.

Note that we should wait on https://github.com/conda-forge/rubinenv-feedstock/pull/277 to merge, which adds the `deepdiff` dependency introduced by this update.

JeremyMcCormick:
Also dropping https://github.com/lsst/felis/pull/109/commits/19213e300e317000f98b937c0d172cef497b44fe and https://github.com/lsst/felis/pull/109/commits/c36134ef48f430292be2bfa0fae78e9e1ceba572

gpdf (approved):
Functionally reviewed in standup today.  Looks good!

PR #119, opened 2024-12-03
DM-47900: Add Python 3.13 support

JeremyMcCormick:
Add Python 3.13 to the build matrix and supported versions.

This was separated from #109 by request of @gpdf.

## Checklist

- [x] Ran Jenkins


timj (approved):
I really don't think this warranted an entire Jira ticket to add two lines but it all looks fine.

PR #120, opened 2024-12-06
DM-47256: Remove apply_schema_to_tables flag

JeremyMcCormick:
This flag is redundant, because enabling `apply_schema_to_metadata` on the `MetaDataBuilder` will already result in the tables being associated to the schema correctly.

## Checklist

- [x] Ran Jenkins


docs/changes/DM-47256.removal
andy-slac, line 1: Does this file need an extension? Looking at the contents I think this should be restructured text, but I'm not sure what towncrier assumptions are?

andy-slac (approved):
Looks good, one minor question.

docs/changes/DM-47256.removal
timj: It will need the .rst extension.

JeremyMcCormick: Good catch - fixed now.

PR #131, opened 2025-01-27
DM-48616: Remove old TAP_SCHEMA module

JeremyMcCormick:
## Checklist

- [x] Ran Jenkins


python/felis/cli.py
```diff
@@ -417,6 +251,40 @@ def load_tap_schema(
     ).load()
 
 
+@cli.command("init-tap-schema", help="Initialize a standard TAP_SCHEMA database")
+@click.option("--engine-url", envvar="FELIS_ENGINE_URL", help="SQLAlchemy Engine URL")
+@click.option("--tap-schema-name", help="Name of the TAP_SCHEMA schema in the database")
```
stvoutsin, line 256: Is it perhaps worth documenting that this defaults to TAP_SCHEMA if left empty?

stvoutsin (approved):
These changes look good to me. I haven't reviewed the `tap_schema` module extensively but as far as the changes to the cli here I can't see any issues.

python/felis/cli.py
JeremyMcCormick: Good idea. I will add a note on this.

PR #132, opened 2025-02-12
DM-48925: Add serialization support

JeremyMcCormick:
## Checklist

- [ ] Ran Jenkins

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -32,7 +32,15 @@ import yaml
 from astropy import units as units  # type: ignore
 from astropy.io.votable import ucd  # type: ignore
 from lsst.resources import ResourcePath, ResourcePathExpression
-from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, field_validator, model_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    ValidationInfo,
+    field_serializer,
+    field_validator,
+    model_validator,
+)
 
 from .db.dialects import get_supported_dialects
 from .db.sqltypes import get_type_func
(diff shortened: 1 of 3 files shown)


python/felis/datamodel.py
```diff
@@ -513,6 +513,11 @@ class CheckConstraint(Constraint):
     expression: str
     """Expression for the check constraint."""
 
+    @field_serializer("type")
+    def serialize_type(self, value: str) -> str:
+        """Ensure '@type' is included in serialized output."""
+        return value
```
TallJimbo, line 522: I don't understand why this method is necessary.  Wouldn't pydantic save it even if it's not present?

tests/test_datamodel.py
```diff
@@ -880,5 +883,43 @@ def test_precision(self) -> None:
...
+        with open(TEST_SERIALIZATION) as file:
+            original_yaml_content = file.read()
+
+        # Load the schema from the original YAML content
+        schema_out = Schema.from_uri(TEST_SERIALIZATION)
+        serialized_data = schema_out.model_dump(by_alias=True, exclude_none=True, exclude_defaults=True)
+
+        # Write the serialized data to a temporary YAML file
+        with tempfile.NamedTemporaryFile(delete=False, suffix=".yaml", mode="w+") as temp_file:
+            print(f"Dumping schema to {temp_file.name}")
```
TallJimbo, line 901: Did you intend to leave the debugging prints in this test?  I don't usually find them that useful, since pytest gathers up the stdout and shows it out of order from other test output.

TallJimbo (approved):

python/felis/datamodel.py
JeremyMcCormick: I don't fully understand it either, but Pydantic was ignoring the `type` field without this.

tests/test_datamodel.py
JeremyMcCormick: I want to print the differences between the original and reread files if there are any, but this particular print out isn't useful and can be removed.

python/felis/datamodel.py
JeremyMcCormick: (As discussed, this is because I'm excluding default values when serializing.)

PR #136, opened 2025-03-10
DM-45619: Add support to tap_schema for generating unique key_id values

JeremyMcCormick:
## Checklist

- [x] Ran Jenkins

### DIFF
--- a/python/felis/tap_schema.py
+++ b/python/felis/tap_schema.py
@@ -27,7 +27,7 @@ import re
 from typing import Any
 
 from lsst.resources import ResourcePath
-from sqlalchemy import MetaData, Table, text
+from sqlalchemy import MetaData, Table, select, text
 from sqlalchemy.engine import Connection, Engine
 from sqlalchemy.engine.mock import MockConnection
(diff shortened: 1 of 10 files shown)


python/felis/tap_schema.py
```diff
@@ -365,6 +365,34 @@ def initialize_database(self, engine: Engine) -> None:
...
+        -------
+        list
+            A list of dictionaries containing the rows from the table.
+        """
+        table = self[table_name]
+        with engine.connect() as connection:
+            if filter_condition:
+                result = connection.execute(select(table).where(text(filter_condition)))
+            else:
+                result = connection.execute(select(table))
```
ktlim, line 392:
```suggestion
        query = select(table)
        if filter_condition:
            query = query.where(text(filter_condition))
        with engine.connect() as connection:
            result = connection.execute(query)
```

ktlim (approved):
Seems reasonable.  The new `select()` method is only for the test?

JeremyMcCormick:
> Seems reasonable. The new `select()` method is only for the test?

Yes, just used in the test for now.

PR #145, opened 2025-05-15
DM-50914: Add flag for forcing unbounded votable arraysize values of '*'

JeremyMcCormick:
## Checklist

- [x] Ran Jenkins


python/felis/cli.py
```diff
@@ -198,6 +198,11 @@ def create(
 @click.option("--dry-run", "-D", is_flag=True, help="Execute dry run only. Does not insert any data.")
 @click.option("--echo", "-e", is_flag=True, help="Print out the generated insert statements to stdout")
 @click.option("--output-file", "-o", type=click.Path(), help="Write SQL commands to a file")
+@click.option(
+    "--force-unbounded-arraysize",
+    is_flag=True,
+    help="Force unbounded array size for all variable length string columns, e.g., ``votable:arraysize: *``",
```
gpdf, line 204: How about "use unbounded arraysize by default for all variable-length string columns",

and a comment in the code that explains why this option is even there at all.

python/felis/datamodel.py
```diff
@@ -452,7 +455,14 @@ def check_votable_arraysize(cls, values: dict[str, Any]) -> dict[str, Any]:
                 if datatype == "char":
                     arraysize = str(length)
                 elif datatype in ("string", "unicode", "binary"):
-                    arraysize = f"{length}*"
+                    if context.get("force_unbounded_arraysize", False):
+                        arraysize = "*"
+                        logger.debug(
+                            f"Forced 'votable:arraysize' to '*' on column '{values['name']}' with datatype "
```
gpdf, line 461:
```suggestion
                            f"Forced VOTable's 'arraysize' to '*' on column '{values['name']}' with datatype "
```

gpdf (commented):
Two small suggestions

PR #152, opened 2025-07-03
DM-51707: Fix the handling of composite keys in the tap_schema module

JeremyMcCormick:
## Checklist

- [ ] Ran Jenkins


gpdf (approved):
This fixes the urgent issue and improves validation.  OK to merge.

I'm not sure it addressed the "user writes a composite key relationship that connects more than two tables" concern, but that was a real edge case and it's fine to merge without that.

PR #161, opened 2025-10-27
DM-53031: Add option to instantiate TAP_SCHEMA with extensions to table columns

stvoutsin:
## Description

As part of the plan to move TAP_SCHEMA to CloudSQL, we are going to be using felis to generate the table/schema structures and to populate the metadata as described in [SQR-107](https://sqr-107.lsst.io/). For this to work we need the additional columns that are expected for the CADC TAP service (e.g. `api_created`). These are non-standard columns, so I've gone with the approach here of introducing an extension parameter that can be used to link to additional columns that can be added to a table. 

Alternatively if we want a more simple approach and don't care about making felis specific to the CADC TAP service we could also add them directly to the existing metadata.

Or also I'm open to other/better solutions that I've not considered (not very familiar with the felis codebase so let me know if there are).

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/python/felis/tap_schema.py
+++ b/python/felis/tap_schema.py
@@ -68,6 +68,8 @@ class TableManager:
         A string to append to all the standard table names.
         This needs to be used in a way such that the resultant table names
         map to tables within the TAP_SCHEMA database.
+    extensions_path
+        The path to additional TAP_SCHEMA table definitions.
 
     Notes
     -----
@@ -88,12 +90,13 @@ class TableManager:
         schema_name: str | None = None,
         apply_schema_to_metadata: bool = True,
         table_name_postfix: str = "",
+        extensions_path: str | None = None,
     ):
         """Initialize the table manager."""
         self.table_name_postfix = table_name_postfix
         self.apply_schema_to_metadata = apply_schema_to_metadata
         self.schema_name = schema_name or TableManager._SCHEMA_NAME_STD
-        self.table_name_postfix = table_name_postfix
+        self.extensions_path = extensions_path
 
         if is_valid_engine(engine):
             assert isinstance(engine, Engine)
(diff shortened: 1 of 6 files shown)


JeremyMcCormick (changes requested):
Overall, I like the approach very much, especially managing the extensions as a Felis YAML file and making them optional so we maintain the core TAP_SCHEMA representation separately.

I requested some changes I would like to see with the implementation before approving. Please see the inline comments.

python/felis/tap_schema.py
```diff
@@ -109,10 +115,68 @@ def __init__(
...
+            if table.name not in ext_data["tables"]:
+                continue
+
+            table_extensions = ext_data["tables"][table.name]
+            if not table_extensions:
+                continue
+
+            extension_columns = []
+            for col_def in table_extensions:
+                if "name" not in col_def or "datatype" not in col_def:
```
JeremyMcCormick: If you load the data into Felis as suggested above then you don't need to check for this, as they are required fields and a validation error would be thrown if they are missing.

python/felis/tap_schema.py
```diff
@@ -109,10 +115,68 @@ def __init__(
...
+                if "@id" not in col_def:
+                    col_def["@id"] = f"#{table.name}.{col_def['name']}"
+
+                ext_column = datamodel.Column(**col_def)
+                extension_columns.append(ext_column)
+                extension_count += 1
+
+            table.columns = list(table.columns) + extension_columns
+
+        self._metadata = MetaDataBuilder(
```
JeremyMcCormick: Probably the code should be rearranged so we don't call this twice.

JeremyMcCormick:
> > Overall, I like the approach very much, especially managing the extensions as a Felis YAML file and making them optional so we maintain the core TAP_SCHEMA representation separately.
> 
> > 
> 
> > I requested some changes I would like to see with the implementation before approving. Please see the inline comments.
> 
> 
> 
> I think I've applied your suggested changes, but let me know if I've missed anything or have misinterpreted any requested change

So you have made it that:

- extensions are on by default
- `--extensions ${uri}` overrides the default
- `--no-extensions` disables extensions 

And I guess `--extensions` and `--no-extensions` are mutually exclusive?

Apologies if there was confusion about what was requested. If you want to just simplify this so that extensions are off by default and then `--extensions ${uri}` turns them on, that would probably be fine. I don't think they are going to be used much outside Rubin so if the URI has to be provided that is probably fine.

JeremyMcCormick:
@stvoutsin Can you deploy this ticket branch to the TAP service on `data-int` with a Phalanx branch  just to make sure nothing breaks inadvertently?

I like to do this when making any changes to the `tap_schema` module, just to be safe. I expect it is fine but would just like to make sure.

PR #166, opened 2025-11-19
DM-53293: Refactor database utility classes

JeremyMcCormick:
This PR heavily refactors the database utility modules in the `db` package:

- The `DatabaseContext` class was moved from the (deleted) `utils` module to its own module and made purely abstract so that it functions as an interface now. Each database type has its own sub-class (SQLite, PostgreSQL, etc.) in which variant-specific operations are implemented such as initialization and drop.
- The modules in the `db` package were significantly consolidated and reorganized.
  - Several modules with only a few functions were removed and these functions were moved into other modules as appropriate.
  - All of the modules in the package were prepended with underscores and removed from the Sphinx docs build, as they do not represent a set of public APIs that should be exposed to users.
- The database operations performed by the CLI in `felis/cli.py` were reworked to use the new modules and classes.
- The `tap_schema` module was changed to use the database context instead of managing metadata and engine independently.
- Tests were updated and added based on the changes; in particular, the `tests/test_database_context.py` tests run some basic database operations on all supported variants.
- MySQL support was added to the main GitHub build workflow.
- A test was added ensuring that all of the modules in the `felis` package can be imported successfully (miscellaneous).

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/docs/changes/DM-53293.misc.rst
+++ b/docs/changes/DM-53293.misc.rst
@@ -1,0 +1,1 @@
+Refactored modules, classes and functions used for database operations


tests/test_database_context.py
```diff
@@ -0,0 +1,141 @@
...
+        self._dialect_name = "mysql"
+        self._engine_url = mysql_engine_url
+
+
+class MockTestCase(DatabaseContextTestCase, unittest.TestCase):
+    """Tests of mock database context."""
+
+    def setUp(self) -> None:
+        super().setUp()
+        # This URL should result in a a mock connection being setup.
```
JeremyMcCormick: fix "a a"

python/felis/db/_database_context.py
```diff
@@ -0,0 +1,594 @@
...
+                    new_params[key] = value
+            print(sql_str % new_params, file=self.file)
+
+
+class DatabaseContextError(Exception):
+    """Exception raised for errors in the DatabaseContext operations."""
+
+
+class DatabaseContext:
+    """Interface for managing database context for different RDMS systems."""
```
JeremyMcCormick: "different database dialects"

python/felis/tap_schema.py
```diff
@@ -84,9 +82,10 @@ class TableManager:
     _SCHEMA_NAME_STD = "TAP_SCHEMA"
     """The standard schema name for the TAP_SCHEMA tables."""
 
+    # Replace engine init with from_engine() method
```
JeremyMcCormick: remove this comment

python/felis/db/_dialects.py
```diff
@@ -114,3 +119,63 @@ def get_dialect_module(dialect_name: str) -> ModuleType:
...
+        type_class = getattr(types, type_name.upper(), None)
+    else:
+        try:
+            dialect_module = get_dialect_module(dialect.name)
+        except KeyError:
+            raise ValueError(f"Unsupported dialect: {dialect}")
+        type_class = getattr(dialect_module, type_name.upper(), None)
+
+    if not type_class:
+        raise ValueError(f"Unsupported type: {type_class}")
```
stvoutsin: Should this be:
`    raise ValueError(f"Unsupported type: {type_name.upper()}")
`

python/felis/cli.py
```diff
@@ -337,19 +307,16 @@ def load_tap_schema(
...
     mgr = TableManager(
-        engine=engine,
-        apply_schema_to_metadata=False if engine.dialect.name == "sqlite" else True,
+        engine_url=engine_url,
         schema_name=tap_schema_name,
         table_name_postfix=tap_tables_postfix,
     )
 
+    # Create DatabaseContext using TableManager's metadata
+    db_ctx = create_database_context(engine_url, mgr.metadata, dry_run=dry_run)
```
stvoutsin: Should this be passing in echo & output_file as well or is this by design?

python/felis/db/database_context.py
```diff
@@ -0,0 +1,859 @@
...
+                        if index.name in existing_indexes:
+                            logger.warning(f"Skipping creation of index '{index.name}' which already exists")
+                            continue
+
+                        index.create(bind=conn, checkfirst=False)  # We already checked
+                        logger.info(f"Created index '{index.name}'")
+            except SQLAlchemyError as e:
+                raise DatabaseContextError("Error creating indexes", e)
+
+    def drop_indexes(self) -> None:
```
stvoutsin: The create/drop indexes methods seem mostly identical, suggestion to extract into a common method, something like `def _manage_index(self, action: str)`
and then they become:

```
def create_indexes(self) -> None:
    self._manage_index("create")

def drop_indexes(self) -> None:
    self._manage_index("drop")
```

python/felis/db/database_context.py
```diff
@@ -0,0 +1,859 @@
...
+_ContextDecorator: TypeAlias = Callable[[_ContextClass], _ContextClass]
+
+
+class DatabaseContextFactory:
+    """Factory for creating DatabaseContext instances based on dialect type."""
+
+    _registry: dict[SupportedDialect, _ContextClass] = {}
+
+    @classmethod
+    def register(cls, dialect: SupportedDialect) -> _ContextDecorator:
```
stvoutsin: I think the idea here is that the registry will only be populated at module import time and should be read-only after (thread-safety) so perhaps I would document that here.

python/felis/db/database_context.py
```diff
@@ -0,0 +1,859 @@
...
+
+        Raises
+        ------
+        DatabaseContextError
+            If there is an error executing the SQL statement.
+        """
+        ...
+
+
+class _BaseContext(DatabaseContext):
```
stvoutsin: This should probably be abstract (ABC) as well?
And if so you'd have to add the abstract methods (initialize & drop)

stvoutsin (approved):
I think the changes looks solid and it's a good approach, some minor suggestions noted here. 
Main question out of those is how idempotency is and should be handled.

JeremyMcCormick:
> I think the changes looks solid and it's a good approach, some minor suggestions noted here. 
> 
> Main question out of those is how idempotency is and should be handled.

Thank you for looking at this. There is also some (relatively minor) duplication of concerns around dialects/variants for handling Felis types which I didn't want to touch right now. I may try to clean that up and consolidate on a subsequent ticket.

python/felis/db/database_context.py
JeremyMcCormick: An inherited abstract class is still considered abstract unless it implements all abstract methods, so I think it will be sufficient just to make `DatabaseContext` an `ABC`.

python/felis/cli.py
JeremyMcCormick: Good catch - it should be passing these arguments. I'll add them.

python/felis/db/database_context.py
```diff
@@ -0,0 +1,896 @@
...
+
+        Parameters
+        ----------
+        dialect
+            The dialect name to register.
+        context_class
+            The context class to use for this dialect.
+        """
+        dialect_name = dialect.lower()
+        if "+" in dialect_name:
```
JeremyMcCormick: Unnecessary since dialect names should not include driver name.

PR #174, opened 2026-01-29
DM-53979: Fix issue with attribute copying from column ref overrides

JeremyMcCormick:
## Checklist

- [x] Ran Jenkins


timj (commented):
How do you know you fixed the bug if there is no test?

timj (approved):
Thanks for explaining that model_dump is the problem here. Please add some context to the commit message stating that the fundamental issue is that you have to avoid model_dump because it coerces types to JSON-safe form.

PR #175, opened 2026-01-30
DM-53999: Improve handling and validation of column overrides

JeremyMcCormick:
This PR improves the handling of column overrides:

- Added an `_update_from_overrides` method on `Column` for updating attributes from a `ColumnOverrides` object
  - Used `model_fields_set` in this method so that explicitly setting some attributes to `None` is allowed (specifically, `description` and `tap_column_index`)
  - Added a validation method to make sure that overriding certain fields to `None` is disallowed
- Set `model_config` on `ColumnOverrides`, copied from the default Felis config (needs to be done explicitly since `ColumnOverrides` does not inherit from Felis's base model class)
- Added significantly better test coverage

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/python/felis/datamodel.py
+++ b/python/felis/datamodel.py
@@ -541,6 +541,25 @@ class Column(BaseObject):
             self.votable_xtype = "timestamp"
         return self
 
+    def _update_from_overrides(self, overrides: ColumnOverrides) -> None:
+        """Update the column attributes from the given overrides.
+
+        Parameters
+        ----------
+        overrides
+            The column overrides to apply or `None` to skip applying overrides.
+
+        Notes
+        -----
+        Using ``model_fields_set`` allows updating only the fields that are
+        explicitly set in the `overrides` object. This prevents overwriting
+        existing column attributes which were not explicitly provided.
+        """
+        if overrides.model_fields_set:
+            logger.debug("Applying overrides to column '%s': %s", self.id, overrides.model_fields_set)
+            for field in overrides.model_fields_set:
+                setattr(self, field, getattr(overrides, field))
+
 
 class Constraint(BaseObject):
     """Table constraint model."""
@@ -819,9 +838,14 @@ class ColumnGroup(BaseObject):
 class ColumnOverrides(BaseModel):
     """Allowed overrides for a referenced column.
 
-    All fields are optional; missing values mean "inherit from the base
-    column".
+    Notes
+    -----
+    All of these fields are optional. Values of None may be explicitly set to
+    override the corresponding attribute in the referenced column but only
+    for certain fields (see validation in `_check_non_nullable_overrides`).
     """
+
+    model_config = CONFIG.copy()
 
     datatype: DataType | None = None
     """New datatype for the column."""
@@ -832,7 +856,7 @@ class ColumnOverrides(BaseModel):
     description: str | None = None
     """New description for the column."""
 
-    nullable: bool = True
+    nullable: bool | None = None
     """New nullable flag for the column."""
 
     tap_principal: int | None = Field(default=None, alias="tap:principal")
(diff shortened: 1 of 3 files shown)


python/felis/datamodel.py
```diff
@@ -541,6 +541,39 @@ def check_votable_xtype(self) -> Column:
             self.votable_xtype = "timestamp"
         return self
 
+    def _update_from_overrides(self, overrides: ColumnOverrides | None) -> None:
+        """Update the column attributes from the given overrides or ignore if
```
timj: I would cut off the sentence after the word `overrides`. Then I would move the notes into the parameter section. The parameter docs are where the discussion of `None` vs `ColumnOverrides` is relevant.

python/felis/datamodel.py
```diff
@@ -541,6 +541,39 @@ def check_votable_xtype(self) -> Column:
...
+        overrides is None.
+
+        Parameters
+        ----------
+        overrides
```
timj, line 549: I see you are in modern sphinx land.

python/felis/datamodel.py
```diff
@@ -541,6 +541,39 @@ def check_votable_xtype(self) -> Column:
...
+            The column overrides to apply.
+
+        Notes
+        -----
+        Values of `None` in the overrides are ignored, as these cannot
+        currently be distinguished from defaults in the Pydantic model.
+        """
+        if overrides is not None:
+            if overrides.datatype is not None:
```
timj: This code is very repetitive. How about something like:

```python
for attr in ("datatype", "length", "description", "nullable", "tap_principal", "tap_column_index"):
    if (override := getattr(overrides, attr)) is not None:
        setattr(self, attr, override)
```

python/felis/datamodel.py
```diff
@@ -1406,10 +1426,11 @@ def _process_column_refs(
                     # written out during serialization
                     column_copy._is_resource_ref = True
 
-                # Apply overrides to the original column definition
-                overrides: ColumnOverrides | None = None
+                # Apply overrides to the referenced column definition
+                overrides = column_ref.overrides if column_ref else None
                 if column_ref is not None:
```
timj: Why is this test needed since we did the test the line above to calculate overrides. Can't you remove it completely and then do

```python
column_copy._update_from_overrides(overrides)
```
without any more checks? The column_ref.overrides below is redoing the line above.

tests/test_datamodel.py
```diff
@@ -1551,5 +1552,97 @@ def test_tap_column_index_with_overrides(self) -> None:
...
+            description="Base column",
+            datatype="char",
+            length=64,
+            nullable=False,
+            tap_principal=1,
+            tap_column_index=10,
+        )
+
+        # Override all allowed fields with different values
+        overrides = ColumnOverrides(
```
timj, line 1590: I think we also need a test that only has a couple of entries in the ColumnOverrides to make sure that some things are left unchanged.

timj (changes requested):
I have some comments that I think need addressing.

python/felis/datamodel.py
JeremyMcCormick: Felis uses this style throughout the codebase.

python/felis/datamodel.py
JeremyMcCormick: It is needed because the `overrides` variable is used later in the method, and `column_ref` may be `None`.

timj: I think it would be clearer if the code said
```python
if overrides is not None:
    column_copy._update_from_overrides(overrides)
```
since otherwise if looks like you are not taking any of the information from the line above into account in this block and redoing it all again.

python/felis/datamodel.py
JeremyMcCormick: I changed it to use a loop over `model_fields_set` (the fields that were explicitly provided in the input data/YAML) so that `None` is a valid value for certain columns.

python/felis/datamodel.py
JeremyMcCormick: I changed the method signature of `_update_from_overrides` so that it does not accept `None`, since this is checked externally already.

tests/test_datamodel.py
JeremyMcCormick: I added this as `test_overrides_subset`.

python/felis/datamodel.py
```diff
@@ -541,6 +541,26 @@ def check_votable_xtype(self) -> Column:
...
+
+        Notes
+        -----
+        Using ``model_fields_set`` allows updating only the fields that are
+        explicitly set in the `overrides` object. This prevents overwriting
+        existing column attributes which were not explicitly provided in the
+        overrides.
+        """
+        if overrides.model_fields_set:
+            logger.debug(f"Applying overrides to column '{self.id}': {overrides.model_fields_set}")
```
timj:
```suggestion
            logger.debug("Applying overrides to column '%s': %s", self.id, overrides.model_fields_set)
```
Do not use f-strings in debug logs.

timj (approved):
one comment about logging

PR #180, opened 2026-05-06
DM-54876: Add option to dump command for sorting output columns alphabetically

JeremyMcCormick:
https://github.com/lsst/felis/pull/180/changes/3a07dd65f1a5b96c5e436c196be3ada0b674f253 fixes the Documenteer build for now. I will convert the site to Documenteer 2.x on another PR.

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -605,16 +605,25 @@ def diff(
     help="Remove any column references from resources and inline the full column definitions in the output",
     default=False,
 )
+@click.option(
+    "--sort-columns/--no-sort-columns",
+    is_flag=True,
+    help="Sort columns alphabetically by name in the output",
+    default=False,
+)
 @click.argument("files", nargs=2, type=click.Path())
 @click.pass_context
 def dump(
     ctx: click.Context,
     strip_ids: bool,
     dereference_resources: bool,
+    sort_columns: bool,
     files: list[str],
 ) -> None:
     if strip_ids:
         logger.info("Stripping IDs from the output schema")
+    if sort_columns:
+        logger.info("Sorting output columns alphabetically")
     if files[1].endswith(".json"):
         format = "json"
     elif files[1].endswith(".yaml"):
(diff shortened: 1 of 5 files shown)


tests/test_cli.py
```diff
@@ -287,6 +287,32 @@ def test_dump_json_with_strip_ids(self) -> None:
             except ValueError:
                 self.fail("Dumped YAML contains forbidden key '@id'")
 
+    @classmethod
+    def _check_columns_sorted(cls, data: dict[str, Any]) -> None:
+        """Check that columns in each table are sorted alphabetically by
+        name.
+        """
+        for table in data.get("tables", []):
```
timj, line 357: Is it worth testing that the content got sorted with the column names? ;-)

python/felis/cli.py
```diff
@@ -605,16 +605,25 @@ def diff(
     help="Remove any column references from resources and inline the full column definitions in the output",
     default=False,
 )
+@click.option(
+    "--sort-columns/--no-sort-columns",
```
timj, line 609: Can you check that this interacts as expected with `--dereference-resources`?

timj (approved):
Looks okay.

tests/test_cli.py
JeremyMcCormick: The column fields should be in the order defined by the Pydantic model based on `model_dump()` behavior. Is that what you mean?

JeremyMcCormick: I don't see how this code could possibly effect the column content though:

    for table in data.get("tables", []):
        table["columns"] = sorted(table.get("columns", []), key=itemgetter("name"))

timj: Oh. So the column sorting is just sorting some names and not sorting the tabular data. If it was moving columns around then testing that the data moved with the column names matters. If it's only sorting names I misunderstood.

PR #181, opened 2026-05-13
DM-54856: Add support for resolving referenced schemas automatically and convert the CLI to use resource paths and URIs

JeremyMcCormick:
This PR updates the resolution of relative resource URIs such as `uri: drp_base.yaml` to use the schema's base path. For instance, if the schema was loaded from `file:///tmp/yml/schema.yaml`, then for a resource this would resolve to `file:///tmp/tmp/drp_base.yaml`. Referenced URIs with absolute paths will not trigger this behavior.

The command line interface was also converted to use URIs, because the previous behavior using file streams did not provide URIs for this resolution.

## Checklist

- [x] Ran Jenkins

### DIFF
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -39,7 +39,7 @@ from .tap_schema import DataLoader, MetadataInserter, TableManager
 
 __all__ = ["cli"]
 
-logger = logging.getLogger("felis")
+logger = logging.getLogger(__name__)
 
 loglevel_choices = ["CRITICAL", "FATAL", "ERROR", "WARNING", "INFO", "DEBUG"]
 
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -102,8 +102,8 @@ def cli(
     ctx.obj["column_ref_index_increment"] = column_ref_index_increment
     if ctx.obj["column_ref_index_increment"] is not None:
         logger.info(
-            f"Automatic indexing of column references is enabled with increment "
-            f"{ctx.obj['column_ref_index_increment']}"
+            "Automatic indexing of column references is enabled with increment %s",
+            ctx.obj["column_ref_index_increment"],
         )
 
 
@@ -492,7 +492,7 @@ def validate(
     """
     rc = 0
     for uri in uris:
-        logger.info(f"Validating schema at '{uri}'")
+        logger.info("Validating schema at '%s'", uri)
         try:
             Schema.from_uri(
                 uri,
@@ -505,7 +505,7 @@ def validate(
                     "column_ref_index_increment": ctx.obj["column_ref_index_increment"],
                 },
             )
-            logger.info(f"Successfully validated schema at '{uri}'")
+            logger.info("Successfully validated schema at '%s'", uri)
         except ValidationError as e:
             logger.error(e)
             rc = 1
@@ -636,7 +636,7 @@ def dump(
             schema.dump_yaml(f, strip_ids=strip_ids, sort_columns=sort_columns)
         elif format == "json":
             schema.dump_json(f, strip_ids=strip_ids, sort_columns=sort_columns)
-    logger.info(f"Dumped '{uris[0]}' to '{uris[1]}'")
+    logger.info("Dumped '%s' to '%s'", uris[0], uris[1])
 
 
 if __name__ == "__main__":
--- a/python/felis/cli.py
+++ b/python/felis/cli.py
@@ -83,11 +83,18 @@ def cli(
     """Felis command line tools"""
     ctx.ensure_object(dict)
 
-    # Configure logging (must come first)
+    # Configure logging on the felis package logger directly.
+    # Do not use basicConfig() as it is a no-op when handlers already exist
+    # (e.g., when running under pytest).
+    felis_logger = logging.getLogger("felis")
+    felis_logger.setLevel(log_level)
     if log_file:
-        logging.basicConfig(filename=log_file, level=log_level)
+        handler: logging.Handler = logging.FileHandler(log_file)
     else:
-        logging.basicConfig(level=log_level)
+        handler = logging.StreamHandler()
+    handler.setLevel(log_level)
+    handler.setFormatter(logging.Formatter("%(levelname)s:%(name)s:%(message)s"))
+    felis_logger.addHandler(handler)
 
     # Configure ID generation (flag can only turn it off)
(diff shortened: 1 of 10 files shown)


python/felis/cli.py
```diff
@@ -533,27 +536,22 @@ def validate(
...
-    files_list = list(files)
-    schemas = [
-        Schema.from_stream(file, context={"id_generation": ctx.obj["id_generation"]}) for file in files_list
-    ]
+    uri_list = list(uris)
+    schemas = [Schema.from_uri(uri, context={"id_generation": ctx.obj["id_generation"]}) for uri in uri_list]
     diff: SchemaDiff
     if len(schemas) == 2:
         if comparator == "alembic":
             # Reset file stream to beginning before re-reading
```
JeremyMcCormick: Delete this comment.

python/felis/datamodel.py
```diff
@@ -1268,12 +1268,12 @@ class Schema(BaseObject, Generic[T]):
     version: SchemaVersion | str | None = None
     """The version of the schema."""
 
-    tables: Sequence[Table]
-    """The tables in the schema."""
-
     resources: dict[str, Resource] = Field(default_factory=dict)
     """External resources referenced by this schema."""
 
+    tables: Sequence[Table]
```
timj, line 1287: Why did this move in the file? Can you put it back so we don't get a diff?

python/felis/datamodel.py
```diff
@@ -1906,10 +1914,14 @@ def from_uri(cls, resource_path: ResourcePathExpression, context: dict[str, Any]
             Raised if the schema fails validation.
         """
         try:
-            rp_stream = ResourcePath(resource_path).read()
+            rp = ResourcePath(resource_path, forceAbsolute=False)
```
timj:
```suggestion
            rp = ResourcePath(resource_path, forceAbsolute=False, forceDirectory=False)
```
Safer to assert that you know it's meant to be a file.

python/felis/datamodel.py
```diff
@@ -1302,8 +1302,16 @@ def _load_resources(self: Schema, info: ValidationInfo) -> Schema:
         else:
             context = {}
 
+        resource_path = context.pop("resource_path", None)
```
timj, line 1320: Maybe do the parenting here instead of in the loop?

```python
base_uri = None
if resource_path is not None:
    base_uri = resource_path.parent()
```

python/felis/datamodel.py
```diff
@@ -1302,8 +1302,16 @@ def _load_resources(self: Schema, info: ValidationInfo) -> Schema:
...
+
         for resource_name, resource in self.resources.items():
             uri = resource.uri
+            if resource_path is not None:
+                uri = resource_path.parent().join(uri)
```
timj:
```suggestion
                uri = base_uri.join(uri, forceDirectory=False)
```

python/felis/datamodel.py
```diff
@@ -1302,8 +1302,16 @@ def _load_resources(self: Schema, info: ValidationInfo) -> Schema:
...
+                logger.info(
+                    f"Resolved resource URI for '{resource_name}' to '{uri}' "
+                    f"using base path '{resource_path.parent()}'"
```
timj:
```suggestion
                    f"using base path '{base_uri}'"
```

python/felis/datamodel.py
```diff
@@ -1302,8 +1302,16 @@ def _load_resources(self: Schema, info: ValidationInfo) -> Schema:
         else:
             context = {}
 
+        resource_path = context.pop("resource_path", None)
+
         for resource_name, resource in self.resources.items():
             uri = resource.uri
+            if resource_path is not None:
+                uri = resource_path.parent().join(uri)
+                logger.info(
```
timj: INFO log messages should use `%s` placeholders, not f-strings.

python/felis/datamodel.py
```diff
@@ -1302,8 +1302,16 @@ def _load_resources(self: Schema, info: ValidationInfo) -> Schema:
...
+                    f"Resolved resource URI for '{resource_name}' to '{uri}' "
```
timj: Technically you do not know if the URI got changed with this logic. You would need to compare the before and after URI and see if it changed.

python/felis/metadata.py
```diff
@@ -433,7 +429,6 @@ def create_metadata(
     MetaData
         The SQLAlchemy metadata object with proper schema handling.
     """
-    schema = Schema.from_stream(felis_file, context={"id_generation": id_generation})
```
timj, line 436: What's the motivation for dropping the Schema load from here rather than changing file to uri?

timj (approved):
Looks okay. Still not sure why you didn't change the file to uri and keep the same helper.

python/felis/datamodel.py
JeremyMcCormick: I did this on purpose with a separate commit so when serializing this section comes before tables and not at the end of the file.

python/felis/metadata.py
JeremyMcCormick: It's a cleaner interface if this accepts a schema rather than path so that validation context can already be applied externally. Otherwise, the context or context settings would have to be passed in as well, which is awkward. If loading the schema raises errors they will also occur outside the function, which is desirable. 

The change is relatively minor as this method is only used in a few places. I checked Repertoire and it is not used there.

JeremyMcCormick:
> Looks okay. Still not sure why you didn't change the file to uri and keep the same helper.

Please see https://github.com/lsst/felis/pull/181#discussion_r3242834500.

python/felis/cli.py
```diff
@@ -487,12 +491,11 @@ def validate(
     optional validations controlled by the Pydantic context.
     """
     rc = 0
-    for file in files:
-        file_name = getattr(file, "name", None)
-        logger.info(f"Validating {file_name}")
+    for uri in uris:
+        logger.info(f"Validating schema at '{uri}'")
```
timj: Oh, also fix the logger.info calls to use `%s` everywhere.

JeremyMcCormick: Okay, I'll do that on another commit. I guess all logger calls might as well use that style?

timj: All logging is meant to use deferred formatting.

timj (approved):
Thanks for fixing the loggers.

PR #189, opened 2026-06-30
DM-51789: Add support for looking up column references by name

JeremyMcCormick:
## Summary

This adds support for looking up column references by name throughout the data model. Previously all in-table references, including from the `primaryKey` field, constraints, and indices, required globally-unique `@id` values. They now accept column names directly. ID-based references are retained as a deprecated fallback that logs a warning, so existing schemas continue to validate and build without modification.

## New lookup methods on `Table`

- `_find_column(ref)` — preferred entry point; resolves by name first, falls back to ID with a deprecation warning, and raises `KeyError` if the reference is ambiguous (matches one column's name and a different column's ID simultaneously)
- `_find_column_by_name(name)` — strict name-only lookup; no fallback, no warning
- `_find_column_by_id(id)` — ID-only lookup; always emits a deprecation warning

## New `ForeignKey` reference style

A new `reference` field on `ForeignKeyConstraint` replaces the old `referencedColumns` list of global IDs. It names the target table and columns directly by name:

```yaml
constraints:
  - name: fk_orders_customer
    "@type": ForeignKey
    columns:
      - customer_id
    reference:
      table: customers
      columns:
        - customer_id
```

The old `referencedColumns` style continues to work but is deprecated.

## Schema-level validation

Three `@model_validator` validators on `Schema` now catch unresolvable column references at schema-load time rather than deferring errors to metadata construction:

- `check_constraints` — validates source columns of `ForeignKey` and `Unique` constraints; for name-based FK `reference` style, validates the referenced table and columns strictly by name with no ID fallback
- `check_indexes` — validates index column references
- `check_primary_key` — validates `primaryKey` column references; implemented at schema level (not table level) so that columns imported via `columnRefs` are fully populated before validation runs

## Updated callers

`metadata.py` and `tap_schema.py` updated to use `_find_column()` for all in-table column resolution and `_find_column_by_name()` for name-based FK targets.

## Documentation

User guide updated with a new "Referencing Columns by Name" section covering the preferred name-based style, the backward-compatible ID fallback, worked YAML examples, and deprecation notes for `primaryKey`, `columnGroups`, constraints, and indexes.

## Tests

New tests cover name-based lookup and ID fallback for all reference contexts (primary keys, constraints, indexes, column groups), including the ambiguous reference error path and composite primary keys. The `test_builder` test in `test_metadata.py` was updated to use `table._find_column()` instead of the ID-only `Schema.__getitem__` for primary key verification.

`tests/data/sales.yaml` and `tests/data/test_composite_keys.yaml` updated to use name-based `primaryKey` as the canonical example of the preferred style.

## Checklist

- [ ] Ran Jenkins

[RFC-1111]: https://rubinobs.atlassian.net/browse/RFC-1111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DM-46055]: https://rubinobs.atlassian.net/browse/DM-46055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -226,3 +226,7 @@ format = "png"
 exclude = [
     "sqlalchemy"
 ]
+
+[tool.pyright]
+include = ["python", "tests"]
+extraPaths = ["python"]


python/felis/datamodel.py
```diff
@@ -1063,20 +1104,73 @@ def _find_column_by_id(self, id: str) -> Column:
 
         Raises
         ------
-        ValueError
+        KeyError
             Raised if the column is not found.
         """
         for column in self.columns:
             if column.id == id:
+                logger.warning(
```
timj: As discussed, it might be better to accumulate the schema warnings (by warning type?) and issue a summary at the end as a single warning, and then use VERBOSE here.

timj (approved):
This code seems fine but I'm really worried about the approach in general.

In https://github.com/lsst/images/pull/84 we have a whole machinery in place for doing schema migrations and this is completely missing in felis. The irony is that felis is a tool for schema management.

Somehow in the future we need a way for felis to understand the schema being used for a given felis file. There are two versions in play: there is the version associated with the fundamental layout of the file as read by felis and there is a version of the user schema represented in that file. Currently neither are versioned and felis has to try to guess what is happening on context. Something for discussion later. I will post this comment on Jira.

python/felis/datamodel.py
JeremyMcCormick: @timj I've implemented one warning message per table in https://github.com/lsst/felis/pull/189/changes/0e6016b761468c49dfc26b253a53d4268d12ad5f.

The messages look like:

```
WARNING:felis.datamodel:Deprecated ID lookup(s) encountered in table IsolatedStarStellarMotions: 
['#IsolatedStarStellarMotions.isolated_star_id', '#IsolatedStarStellarMotions.referenceId',
 '#IsolatedStarStellarMotions.isolated_star_id'] - see https://felis.lsst.io/user-guide/model.html#referencing-columns-by-name for name-based referencing style
```

This will still be somewhat noisy for schemas with many tables but accumulating all of the warnings across the entire schema and printing them in a batch is probably beyond the scope of this PR as it would require some substantial changes to the codebase.

